PR #1, opened 2018-07-26
Create content-data-admin app

chao-xian:
Initial setup of app, prior to provisioning on the GOV.UK stack.

Following initial steps from https://docs.publishing.service.gov.uk/manual/setting-up-new-rails-app.html


chao-xian:
[(5) Provision and deploy a new barebones Admin app](https://trello.com/c/kOuisVYY/547-5-provision-and-deploy-a-new-barebones-admin-app)

PR #4, opened 2018-08-02
Set log level on production to info

chao-xian:
This should not be set to debug. There may be a problem with
govuk_app_config, but manually updating for now.


chao-xian:
[(8) Provision and deploy a new barebones Admin app](https://trello.com/c/kOuisVYY/547-8-provision-and-deploy-a-new-barebones-admin-app)

PR #5, opened 2018-08-10
Jenkins configuration

peteglondon:
Add a dummy test to test CI integration.

Fix existing lint errors.
Add a Jenkinsfile.

Trello: https://trello.com/c/kOuisVYY/547-8-provision-and-deploy-a-new-barebones-admin-app


lib/tasks/lint.rake
```diff
@@ -1,5 +1,5 @@
 desc "Run govuk-lint on all files"
 task "lint" do
-  sh "govuk-lint-ruby --format clang"
+  sh "govuk-lint-ruby app config lib spec --format clang"
```
erino: What about adding rails linting as per https://github.com/alphagov/content-publisher/blob/master/lib/tasks/lint.rake#L5

peteglondon: Will make the change now.

Jenkinsfile
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env groovy
+
+library("govuk")
+
+node("postgresql-9.3") {
+  govuk.buildProject(
+    beforeTest: {
+      govuk.setEnvar("TEST_COVERAGE", "true")
+    }
```
erino: What about adding testing with brakeman here:
https://github.com/alphagov/content-publisher/blob/7ce30f12d9878def9254407bf0c707b07af4b810/Jenkinsfile#L13

peteglondon: This is the default for a rails app. as documented here:  https://github.com/alphagov/govuk-jenkinslib - they are running - you can see output in Jenkins.

erino: Ah, looks like it's only recently default to `true` https://github.com/alphagov/govuk-jenkinslib/pull/19

PR #9, opened 2018-08-15
Chart component

maxgds:
This is a frontend component that implements the design system
tabs and table alongside a google chart. It is intended for
representing time based data primarily pictorially but with
a tabular backup to support accessibility and allow an alternative
view of the figures.

If running the app for review, see implementions on the page at /dev as the component guide currently doesn't work properly.


app/views/layouts/application.html.erb
```diff
@@ -11,6 +11,9 @@
       { text: "Log out", href: gds_sign_out_path }
     ]
   }%>
+<%= javascript_include_tag "https://www.gstatic.com/charts/loader.js" %>
+<%= javascript_include_tag "application" %>
+<%= stylesheet_link_tag "application" %>
```
alex-ju: From govuk_publishing_components v9.13 it is possible to move this in the `<head>` of the html

app/assets/javascripts/application.js
```diff
@@ -0,0 +1,3 @@
+//= require govuk_publishing_components/all_components
+//= require chartkick
```
alex-ju: If only the chart component uses chartkick (and there are no plans for other components to use it, would be good to move the require in `chart.js`

app/assets/stylesheets/components/_chart.scss
```diff
@@ -0,0 +1,18 @@
+// Include our GOV.UK components
+@import "govuk_publishing_components/all_components";
```
alex-ju: I don't think we need this import here as it already happens in `application.scss`. If we're doing it to show the dependency on other components (e.g. tabs) would be good to be explicit about it (@import "govuk_publishing_components/components/tabs")

app/assets/stylesheets/application.scss
```diff
@@ -13,3 +13,8 @@
  *= require_tree .
  *= require_self
  */
+
+@import "components/chart.scss";
```
alex-ju: I would import gem components and then local components/styles, to be able to reuse variables or overwrite things if necessary.

app/views/development/index.html.erb
```diff
@@ -1 +1,56 @@
-My page
+<%= content_for :title, "Dev page" %>
+
+<span class="govuk-caption-xl">Data analytics for</span>
+<h1 class="govuk-heading-xl">The page being analysed</h1>
+
+<%= render "govuk_publishing_components/components/govspeak" do %>
```
alex-ju: I'm not sure we need govspeak here, unless we have a lot of content. If it's for styling the paragraph we can use `<p class="govuk-body">`

app/views/development/index.html.erb
```diff
@@ -1 +1,56 @@
...
+  <p>
+    Metrics were last updated yesterday at 00:00.
+  </p>
+<% end %>
+
+<%
+  # This data will come from models/controllers/the api eventually, hard coded for now to produce working views.
```
alex-ju: 👍

config/initializers/assets.rb
```diff
@@ -11,4 +11,4 @@
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in the app/assets
 # folder are already added.
-# Rails.application.config.assets.precompile += %w( admin.js admin.css )
+Rails.application.config.assets.precompile += %w(admin.js admin.css)
```
alex-ju: I think we don't need this line. We can keep it as a comment.

spec/components/chart_spec.rb
```diff
@@ -0,0 +1,96 @@
+require 'rails_helper'
+
+data = {
+  caption: "Unique page views",
+  h_axis_title: "Date",
```
pmanrubia: I have noticed that in `data` there are attributes  are not tested. I would try to keep the fixture size as close as possible to the tests. I would remove the fields that are not tested, or just add a test for them if needed.

spec/components/chart_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+      values: [3, 8, 37, 435, 78, 4, 9, 61, 110, 12, 21, 121]
+    }
+  ]
+}
+describe "Chart", type: :view do
+  def render_component(locals)
+    render partial: "components/#{component_name}", locals: locals
+  end
+
+  def component_name
```
pmanrubia: [It feels better to use `let`](https://relishapp.com/rspec/rspec-core/v/3-8/docs/helper-methods/let-and-let)

spec/components/chart_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    "chart"
+  end
+
+  it "does not render when no data is given" do
+    assert_empty render_component({})
+  end
+
+  it "does not render if keys are missing" do
+    d = data.dup
```
pmanrubia: If you use let, there is no need to do a `dup`.

spec/components/chart_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+      values: [5, 119, 74, 117, 50, 119, 61, 110, 12, 21, 121, 67]
+    },
+    {
+      label: "2018",
+      values: [3, 8, 37, 435, 78, 4, 9, 61, 110, 12, 21, 121]
+    }
+  ]
+}
+describe "Chart", type: :view do
+  def render_component(locals)
```
pmanrubia: You could move this method to the end of the `describe` block. It is a common practice in `RSpec`

app/views/development/index.html.erb
maxgds: The dev page isn't a final page, it's only included here for visibility of the component. I'll be using it as a testing ground as I build so this area will probably expand, so I'll not address this.

app/assets/stylesheets/components/_chart.scss
maxgds: Without this, the line 
@include govuk-visually-hidden;
blows up. 
I can switch it out for @import "../../../node_modules/govuk-frontend/all"; but I'm not sure that's any better? I don't really fully understand how the imports and css dependencies work.

maxgds:
@pmanrubia @alex-ju I believe I've addressed everything you both raised so it would be good to get an ok to merge on this.

I'm not sure why I'm getting errors on build - the file it's highlighting as having issues matches the one in master now.

spec/components/chart_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    assert_select "td:last", text: "121"
+  end
+
+  it "calls the chart library" do
+    render_component(data)
+    assert_select "script", text: /new Chartkick.LineChart/
+  end
+
+  def render_component(locals)
+    render partial: "components/#{component_name}", locals: locals
```
pmanrubia: Might be worth just inlining the `component_name`

spec/rails_helper.rb~Chart component
```diff
@@ -0,0 +1,57 @@
+# This file is copied to spec/ when you run 'rails generate rspec:install'
```
pmanrubia: This file might not be needed

PR #11, opened 2018-08-15
Display an aggregated metric (unique page views) for a single content item

KyleMacPherson:
Display an aggregated metric (unique page views) for a single content item

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,1 +1,1 @@
---require spec_helper
+--require rails_helper
--- a/startup.sh
+++ b/startup.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+bundle check || bundle install
+
+bundle exec rails s -b 0.0.0.0 -p 3230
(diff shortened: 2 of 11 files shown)


startup.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
pmanrubia: I would have created this commit in a separate pull request, because it doesn't feel related to this story.
No need to change it now, but might be worth considering for incoming PRs

.rspec
```diff
@@ -1 +1 @@
---require spec_helper
+--require rails_helper
```
pmanrubia: I would have created this commit in a separate pull request, because it doesn't feel related to this story.
No need to change it now, but might be worth considering for incoming PRs

KyleMacPherson: @pmanrubia good point, will do this next time.

KyleMacPherson:
Have made the changes to commits as requested!

PR #15, opened 2018-08-22
Run bundle update

cbaines:


cbaines:
> We need to hold off the Rails bump as it relates to a recent incident with CPM and this may have some issues too.

Sure, I'll close this and take a look at the dependabot PR's instead.

PR #24, opened 2018-08-23
Render metric timeseries data

KyleMacPherson:
Format and render the time series data that we request from the cpm api
in order to populate the chart and table components.

This is the first iteration so the designs are not final but provides all the vital
data needed for these components.

### DIFF
--- a/app/presenters/chart_presenter.rb
+++ b/app/presenters/chart_presenter.rb
@@ -1,0 +1,48 @@
+class ChartPresenter
+  attr_reader :json
+
+  def initialize(json)
+    @json = json.with_indifferent_access
+  end
+
+  def metric
+    json.keys.first
+  end
+
+  def from
+    json.values.flatten.first[:date]
+  end
+
+  def to
+    json.values.flatten.last[:date]
+  end
+
+  def human_friendly_metric
+    metric.tr('_', ' ').capitalize
+  end
+
+  def chart_data
+    {
+      caption: "#{human_friendly_metric} from #{from.to_date} to #{to.to_date}",
+      chart_label: "#{human_friendly_metric} chart",
+      table_label: "#{human_friendly_metric} table",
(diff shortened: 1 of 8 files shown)


chao-xian:
[(3) Display a chart (unique page views) for a single content item](https://trello.com/c/zms3rqZ1/590-3-display-a-chart-unique-page-views-for-a-single-content-item)

app/services/metrics_service.rb
```diff
@@ -7,12 +7,52 @@ def fetch(base_path:, from:, to:, metric:)
...
+  def keys(data, metric)
+    dates = data[metric].map { |hash| hash['date'] }
+    dates.map { |date| date.last(5) }
+  end
+
+  def values(data, metric)
+    data[metric].map { |hash| hash['value'] }
+  end
+
+  def human_friendly_metric(metric)
```
chao-xian: These private methods can't be tested directly. Another benefit of abstracting these as part of the refactor to presenter classes is that we'll be able to unit test them.

app/presenters/chart_presenter.rb
```diff
@@ -0,0 +1,48 @@
+class ChartPresenter
+  attr_reader :response
```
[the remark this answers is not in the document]
pmanrubia: Could you please add a unit test for this class?

spec/features/single_content_item_spec.rb
```diff
@@ -4,14 +4,36 @@
...
   end
 
-  it 'renders the metric for unique_pageviews' do
+  before do
     content_api_has_metric('base/path',
                            'unique_pageviews',
                            '2000-01-01',
                            '2050-01-01',
                            unique_pageviews: { total: 145_000 })
+    content_api_has_timeseries('base/path',
```
pmanrubia: Sorry I didn't pick this up before, but can you rename `content_api_has_timeseries` and `content_api_has_metric` to `content_data_api_has*`?

PR #30, opened 2018-08-29
Multi chart endpoint

KyleMacPherson:
To reduce the number of requests for fetching multiple metrics,
we are sending metrics as a single query parameter with multiple
values.

This change is reflected in the Content Data API
(content-performance-manager) as well[0].

We also add a gds-api-adapter stub to be more closely
aligned to the way the code would be structured in the gem, allowing
code that does not belong in the MetricsService to be abstracted
into GdsApi:ContentDataApi.

[0] https://github.com/alphagov/content-performance-manager/pull/872

### DIFF
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -7,13 +7,14 @@ class MetricsController < ApplicationController
       base_path: params[:base_path],
       from: params[:from],
       to: params[:to],
-      metric: params[:metric]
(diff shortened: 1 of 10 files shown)


chao-xian:
[(5) Update API to support multiple metrics per basepath (EPIC 2)](https://trello.com/c/vClxSNYC/617-5-update-api-to-support-multiple-metrics-per-basepath-epic-2)

PR #43, opened 2018-09-06
Fix error when the backend has no data

peteglondon:
Currently the app raises an error when no
time series data is returned from the back-end.

This commit checks foe empty series before
attempting to render the chart.


peteglondon:
[(3) Add metrics Charts for the Single Performance Page (EPIC 2)](https://trello.com/c/xZb8cKyy/625-3-add-metrics-charts-for-the-single-performance-page-epic-2)

PR #48, opened 2018-09-10
Handle a GdsApi::HTTPNotFound when getting data

peteglondon:
Currently the app fails with a 500 error
when we get a 404 from the content-data-api.

This commit traps the error and renders the 404 page.

### DIFF
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -1,4 +1,6 @@
 class MetricsController < ApplicationController
+  DEFAULT_METRICS = %w[pageviews unique_pageviews number_of_internal_searches].freeze
+
   def show
     service = MetricsService.new
 
@@ -6,11 +8,15 @@ class MetricsController < ApplicationController
       base_path: params[:base_path],
       from: params[:from],
       to: params[:to],
-      metrics: params[:metrics]
+      metrics: DEFAULT_METRICS
     }
 
     @summary = SingleContentItemPresenter
       .parse_metrics(service.fetch(service_params))
       .parse_time_series(service.fetch_time_series(service_params))
+  end
+
+  rescue_from GdsApi::HTTPNotFound do
+    render file: Rails.root.join('public', '404.html'), status: :not_found
   end
 end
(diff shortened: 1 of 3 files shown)


peteglondon:
[(3) Add metrics Charts for the Single Performance Page (EPIC 2)](https://trello.com/c/xZb8cKyy/625-3-add-metrics-charts-for-the-single-performance-page-epic-2)

app/controllers/metrics_controller.rb
```diff
@@ -12,5 +12,7 @@ def show
     @summary = SingleContentItemPresenter
       .parse_metrics(service.fetch(service_params))
       .parse_time_series(service.fetch_time_series(service_params))
+  rescue GdsApi::HTTPNotFound
```
pmanrubia: What about being declarative about the exceptions with [rescue_from](https://apidock.com/rails/ActiveSupport/Rescuable/ClassMethods/rescue_from) in this controller?. It decreases the size of the action and let the reader focus on what the action does.

spec/support/content_data_api.rb
```diff
@@ -10,6 +10,12 @@ def content_data_api_has_metric(base_path:, from:, to:, metrics:, payload:)
         stub_request(:get, url).to_return(status: 200, body: body)
       end
 
+      def content_data_api_does_not_have_metric(base_path:, from:, to:, metrics:)
+        query = GdsApi::ContentDataApi.new.query(from: from, to: to, metrics: metrics)
```
pmanrubia: Not sure if the parameters: `from`, `to` and `metrics` should be in the signature of this method, because not having metrics between those dates would not imply having the error. The `404` should only happen if the `base_path` is not present. 
I would set up the fixture with:

```ruby
content_data_api_does_not_have_metric(base_path: 'base/path')
```

peteglondon: You need them in there to construct the correct URL to stub, I'm not sure of another way to stub this - any ideas gratefully received.

peteglondon: Renamed the method to `content_data_api_does_not_have_base_path`

PR #51, opened 2018-09-11
Add date range picker

KyleMacPherson:
Add date range picker to single content item page in order to let end users get metrics for the time period in which they are interested in. Use `last 30 days` as default time period if none is specified.

No styling on the radio buttons as a component will be coming in the future which handles this.

Change the id's of charts and tables to make them easier to query. The date is not an important part of the chart because there will only be one per query and if we display two different time series on the same chart it could be difficult to handle.

### DIFF
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -1,20 +1,16 @@
 class MetricsController < ApplicationController
   def show
     service = MetricsService.new
+    date_range = DateRange.new(params[:date_range])
 
     service_params = {
       base_path: params[:base_path],
-      from: params[:from],
-      to: params[:to]
+      date_range: date_range,
     }
 
-    @summary = SingleContentItemPresenter
-      .parse_metrics(
-        metrics: service.fetch_aggregated_data(service_params),
-        from: params[:from],
-        to: params[:to]
-      )
-      .parse_time_series(service.fetch_time_series(service_params))
+    metrics = service.fetch_aggregated_data(service_params)
+    time_series = service.fetch_time_series(service_params)
+    @summary = SingleContentItemPresenter.new(metrics, time_series, date_range)
   end
 
   rescue_from GdsApi::HTTPNotFound do
--- a/app/presenters/single_content_item_presenter.rb
+++ b/app/presenters/single_content_item_presenter.rb
@@ -1,18 +1,19 @@
 class SingleContentItemPresenter
-  attr_reader :unique_pageviews, :unique_pageviews_series,
-    :pageviews, :pageviews_series,
-    :number_of_internal_searches, :number_of_internal_searches_series,
-    :satisfaction_score, :satisfaction_score_series,
-    :metadata, :title
+  attr_reader :unique_pageviews, :pageviews, :unique_pageviews_series,
+    :pageviews_series, :base_path, :title, :published_at, :last_updated,
+    :publishing_organisation, :document_type, :number_of_internal_searches,
+    :number_of_internal_searches_series, :satisfaction_score, :satisfaction_score_series,
+    :date_range, :metadata
 
-  def initialize(from, to)
-    @from = from
-    @to = to
+  def initialize(metrics, time_series, date_range)
+    @date_range = date_range
+    parse_metrics(metrics.with_indifferent_access)
+    parse_time_series(time_series.with_indifferent_access)
   end
 
-  def self.parse_metrics(metrics:, from:, to:)
-    new(from, to).parse_metrics(metrics.deep_symbolize_keys)
-  end
+private
+
+  attr_reader :from, :to
 
   def parse_metrics(metrics)
     @unique_pageviews = metrics[:unique_pageviews]
@@ -27,7 +28,6 @@ class SingleContentItemPresenter
       last_updated: format_date(metrics[:public_updated_at]),
       publishing_organisation: metrics[:primary_organisation_title],
     }
-    self
   end
 
   def parse_time_series(time_series)
@@ -35,15 +35,10 @@ class SingleContentItemPresenter
     @pageviews_series = get_chart_presenter(time_series, :pageviews)
     @number_of_internal_searches_series = get_chart_presenter(time_series, :number_of_internal_searches)
     @satisfaction_score_series = get_chart_presenter(time_series, :satisfaction_score)
-    self
   end
 
-private
-
-  attr_reader :from, :to
-
   def get_chart_presenter(time_series, metric)
-    ChartPresenter.new(json: time_series, metric: metric, from: from, to: to)
+    ChartPresenter.new(json: time_series, metric: metric, from: date_range.from, to: date_range.to)
   end
 
   def format_date(date_str)
--- a/spec/features/date_selection_spec.rb
+++ b/spec/features/date_selection_spec.rb
@@ -1,0 +1,136 @@
+RSpec.describe 'date selection', type: :feature do
+  include GdsApi::TestHelpers::ContentDataApi
+  include MetricsChartSpecHelpers
+  let(:metrics) { %w[pageviews unique_pageviews number_of_internal_searches satisfaction_score] }
+
+  before do
+    initial_page_stub
+  end
+
+  it 'renders data for the last 30 days if no date period is selected' do
+    date_range = build(:date_range, :last_30_days)
+    visit '/metrics/base/path'
+    expect_default_date_period_metrics_to_be_displayed(date_range.from.to_date, date_range.to.to_date)
+  end
+
+  it 'renders data for the last 30 days when `last 30 days` is selected' do
+    date_range = build(:date_range, :last_30_days)
+    stub_response_for_date_range(date_range.from.to_date, date_range.to.to_date)
+    visit_page_and_filter_by_date_range('last-30-days')
+    expect_metrics_for_each_date_to_be_correct(date_range.from.to_date, date_range.to.to_date)
+  end
+
+  it 'renders data for the previous month when `last month` is selected' do
+    date_range = build(:date_range, :last_month)
+    stub_response_for_date_range(date_range.from.to_date, date_range.to.to_date)
+    visit_page_and_filter_by_date_range('last-month')
+    expect_metrics_for_each_date_to_be_correct(date_range.from.to_date, date_range.to.to_date)
+  end
+
+  it 'renders data for the last 3 months when `last 3 months` is selected' do
+    date_range = build(:date_range, :last_3_months)
+    stub_response_for_date_range(date_range.from.to_date, date_range.to.to_date)
+    visit_page_and_filter_by_date_range('last-3-months')
+    expect_metrics_for_each_date_to_be_correct(date_range.from.to_date, date_range.to.to_date)
+  end
+
+  it 'renders data for the last 6 months when `last6 months` is selected' do
+    date_range = build(:date_range, :last_6_months)
+    stub_response_for_date_range(date_range.from.to_date, date_range.to.to_date)
+    visit_page_and_filter_by_date_range('last-6-months')
+    expect_metrics_for_each_date_to_be_correct(date_range.from.to_date, date_range.to.to_date)
+  end
+
+  it 'renders data for the last year when `last year` is selected' do
+    date_range = build(:date_range, :last_1_year)
+    stub_response_for_date_range(date_range.from.to_date, date_range.to.to_date)
+    visit_page_and_filter_by_date_range('last-1-year')
+    expect_metrics_for_each_date_to_be_correct(date_range.from.to_date, date_range.to.to_date)
+  end
+
+  it 'renders data for the last 2 years when `last 2 years` is selected' do
+    date_range = build(:date_range, :last_2_years)
+    stub_response_for_date_range(date_range.from.to_date, date_range.to.to_date)
+    visit_page_and_filter_by_date_range('last-2-years')
+    expect_metrics_for_each_date_to_be_correct(date_range.from.to_date, date_range.to.to_date)
+  end
+
+  def initial_page_stub
+    from = Time.zone.today - 30.days
+    to = Time.zone.today
+    content_data_api_has_metric(base_path: 'base/path',
+                                from: from,
+                                to: to,
+                                metrics: metrics)
+    content_data_api_has_timeseries(base_path: 'base/path',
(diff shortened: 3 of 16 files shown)


KyleMacPherson:
[(3) Add date range section to the Single Performance Page (EPIC 2)](https://trello.com/c/poxmPaeb/622-3-add-date-range-section-to-the-single-performance-page-epic-2)

app/controllers/concerns/date_range_parser.rb
```diff
@@ -0,0 +1,48 @@
+require 'active_support/concern'
+
+module Concerns::DateRangeParser
+  extend ActiveSupport::Concern
+
+  included do
+    def parse_date_range(date_range)
+      date_to_range(date_range)
```
pmanrubia: This method returns a hash with two attributes. A `DateRange` seems like a very useful class for our Domain. What extracting a class `/model/date_range`?

app/controllers/concerns/date_range_parser.rb
```diff
@@ -0,0 +1,48 @@
...
+      {
+        from: (Time.zone.today - 1.year).to_s,
+        to: Time.zone.today.to_s
+      }
+    when 'last-2-years'
+      {
+        from: (Time.zone.today - 2.years).to_s,
+        to: Time.zone.today.to_s
+      }
+    else
```
pmanrubia: Not sure about the `else` statement. If the parameter value is not one of the expected ones, shouldn't we just fail?

app/controllers/concerns/date_range_parser.rb
```diff
@@ -0,0 +1,48 @@
+require 'active_support/concern'
+
+module Concerns::DateRangeParser
```
pmanrubia: We would need a unit test for this class.

app/controllers/metrics_controller.rb
```diff
@@ -1,19 +1,26 @@
...
+    date_range = parse_date_range(params[:date_range])
+    from = date_range[:from]
+    to = date_range[:to]
+
     service_params = {
       base_path: params[:base_path],
-      from: params[:from],
-      to: params[:to],
+      from: from,
+      to: to,
```
pmanrubia: If we extract a DateRange, we can update the service parameters?

app/controllers/metrics_controller.rb
```diff
@@ -1,19 +1,26 @@
...
       metrics: DEFAULT_METRICS
     }
 
     @summary = SingleContentItemPresenter
       .parse_metrics(service.fetch(service_params))
```
pmanrubia: I know this is not part of your codebase, but the methods: `parse_metrics`, `parse_time_series`, `set_date_range` doesn't feel like they belong to the `SingleContentItemPresenter`. They feel like internal behaviour that could run after the initialization (after construction phase)

We should be able to rewrite lines 20-23 into:

```ruby
SingleContentItemPresenter.new(base_path: ---, date_range: --- )
```

app/presenters/single_content_item_presenter.rb
```diff
@@ -25,6 +25,11 @@ def parse_time_series(time_series)
     self
   end
 
+  def set_date_range(date_range)
```
pmanrubia: See my comment above, but I would remove from the public API of this class the methods:
- set_date_range
- parse_metrics
- parse_time_series

spec/presenters/chart_presenter_spec.rb
```diff
@@ -31,7 +31,7 @@
   def unique_pageviews_chart_data
     {
       caption: "Unique pageviews from 2018-01-13 to 2018-01-15",
-      chart_id: "unique_pageviews_2018-01-13-2018-01-15_chart",
+      chart_id: "unique_pageviews_chart",
```
pmanrubia: Could you explain why we have this change in this commit. Seems out of place.

spec/features/date_selection_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+    from = Time.zone.today - 30.days
+    visit '/metrics/base/path?metrics[]=unique_pageviews&metrics[]=pageviews'
+
+    expect_default_date_period_metrics_to_be_displayed(from, to)
+  end
+
+  it 'renders data for the last 30 days when `last 30 days` is selected' do
+    to = Time.zone.today
+    from = Time.zone.today - 30.days
+    stub_for(from, to)
```
pmanrubia: Can you rename the method `stub_for` to something more descriptive?

spec/features/date_selection_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+  it 'renders data for the last 30 days if no date period is selected' do
+    to = Time.zone.today
+    from = Time.zone.today - 30.days
+    visit '/metrics/base/path?metrics[]=unique_pageviews&metrics[]=pageviews'
+
+    expect_default_date_period_metrics_to_be_displayed(from, to)
+  end
+
+  it 'renders data for the last 30 days when `last 30 days` is selected' do
+    to = Time.zone.today
```
pmanrubia: Can you use a DateRange factory, using FactoryBot? You could use traits for it. 
This will dry-up the tests and will set a foundation for other tests to reuse it.

spec/features/date_selection_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+    from = Time.zone.today - 30.days
+    stub_for(from, to)
+
+    visit '/metrics/base/path?metrics[]=unique_pageviews&metrics[]=pageviews'
+    find('#date_range_last-30-days').click
+    click_button 'OK'
+    sleep 2
```
pmanrubia: I would get rid of the `sleep 2` by using an assertion:

```ruby
expect(page).to have_selector('selector to the area where the click button lives')
```

This way we will be waiting for the button to appear, and as a side effect, we are also sure that the section exists.

spec/features/date_selection_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+
+    expect_metrics_for_each_date_to_be_correct(from, to)
+  end
+
+  it 'renders data for the previous month when `last month` is selected' do
+    to = Time.zone.today.last_month.end_of_month
+    from = Time.zone.today.last_month.beginning_of_month
+    stub_for(from, to)
+
+    visit '/metrics/base/path?metrics[]=unique_pageviews&metrics[]=pageviews'
```
pmanrubia: You could extract a method for lines 35-39 like:

```ruby
def when_i_filter_by_date_range('last-month')
#
#
#
end
```

spec/features/date_selection_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+
+  def initial_page_stub
+    from = Time.zone.today - 30.days
+    to = Time.zone.today
+
+    content_data_api_has_metric(base_path: 'base/path',
+      from: from,
+      to: to,
+      metrics: %w[unique_pageviews pageviews number_of_internal_searches],
+      payload: {
```
pmanrubia: Do we need to be so explicit with the `payload`?
It should be enough with setting a few attributes (like `base_path`) and default the rest within `content_data_api_has_metric`?

spec/features/date_selection_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+        first_published_at: '2018-02-01T00:00:00.000Z',
+        public_updated_at: '2018-04-25T00:00:00.000Z',
+        primary_organisation_title: 'The ministry',
+        document_type: "news_story"
+      })
+
+    content_data_api_has_timeseries(base_path: 'base/path',
+      from: from,
+      to: to,
+      metrics: %w[unique_pageviews pageviews number_of_internal_searches],
```
pmanrubia: Is it ok to have three metrics, and return only two sets of series?

spec/features/date_selection_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+  end
+
+  def expect_metrics_for_each_date_to_be_correct(from, to)
+    date1 = (from - 1.day).to_s.last(5)
+    date2 = (from - 2.days).to_s.last(5)
+    date3 = (to + 1.day).to_s.last(5)
+
+    unique_pageviews_rows = find("#unique_pageviews_table").all('tr')
+
+    expect(unique_pageviews_rows.count).to eq 4
```
pmanrubia: Can you please use `match_array` from RSpec? This will dry it up.

spec/features/date_selection_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+        pageviews: [
+          { "date" => (from - 1.day).to_s, "value" => 10 },
+          { "date" => (from - 2.days).to_s, "value" => 20 },
+          { "date" => (to + 1.day).to_s, "value" => 30 }
+        ]
+      })
+  end
+
+  def expect_metrics_for_each_date_to_be_correct(from, to)
+    date1 = (from - 1.day).to_s.last(5)
```
pmanrubia: By reading the code it is not obvious (to me, at least) what it does, or what is the shape of date1. Can you think of a better way to describe it?

app/controllers/concerns/date_range_parser.rb
KyleMacPherson: When we first render a single item page I believe we are defaulting to 30 days i.e before anyone has had the chance to pick a date range

spec/presenters/chart_presenter_spec.rb
KyleMacPherson: Yeah, I made a note of it above: 

`Change the id's of charts and tables to make them easier to query. The date is not an important part of the chart because there will only be one per query and if we display two different time series on the same chart it could be difficult to handle`

spec/features/date_selection_spec.rb
KyleMacPherson: It is OK as far as the tests are concerned as we stub the response but I have changed the payload to more accurately reflect what real life response

spec/features/date_selection_spec.rb
KyleMacPherson: Done!

spec/features/date_selection_spec.rb
KyleMacPherson: Done!

spec/features/date_selection_spec.rb
KyleMacPherson: Done! Also added `visit ...` to this method.

spec/features/date_selection_spec.rb
KyleMacPherson: The sleeps are not necessary, forgot to remove a couple sorry! 🤦‍♂️

app/controllers/concerns/date_range_parser.rb
pmanrubia: Extracting a class like DateRange would have the following benefits in the short term:

1. The concern does not return a Hash that you later on have to inspect to extract `to` and `from`
2. We will have to implement shortly retrieving the metrics for the month before, so we can always add a method to perform this operation

We would be able to move from:

```ruby
    date_range = parse_date_range(params[:date_range])
    from = date_range[:from]
    to = date_range[:to]

    service_params = {
      base_path: params[:base_path],
      from: from,
      to: to,
      metrics: DEFAULT_METRICS
    }
```

to 

```ruby
  service_params = {
   base_path: params[:base_path],
   date_range: parse_date_range(params),
   metrics: DEFAULT_METRICS
  } 
```

app/controllers/concerns/date_range_parser.rb
pmanrubia: If that is the case, shouldn't that option be the default checkbox selection? So, the filter is always consistent with the result of the list? This might be something to discuss about with Jeremy/Polly.

app/controllers/metrics_controller.rb
pmanrubia: The `date_range` has `to` and `from` attributes. Not sure why we would need to change it?

spec/presenters/chart_presenter_spec.rb
pmanrubia: Sorry, I forgot about your PR description. Do you think It would have been possible to isolate this change in a commit and add those notes in the commit?

spec/features/date_selection_spec.rb
pmanrubia: Sorry, if I wasn't clear enough. I meant having to be so explicit in the spec. We are not testing any of the metadata, so they shouldn't be part of the Spec. We could hide those attributes in the helper method, and override them with the values passed as parameters.

spec/presenters/chart_presenter_spec.rb
KyleMacPherson: Sure, done!

spec/features/date_selection_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+  end
+
+  def initial_page_stub
+    from = Time.zone.today - 30.days
+    to = Time.zone.today
+    content_data_api_has_metric(base_path: 'base/path',
+                                from: from,
+                                to: to,
+                                metrics: metrics)
+    content_data_api_has_timeseries(base_path: 'base/path',
```
pmanrubia: It seems to me that the `initial_page_stub` is a way to populate some data so we 
stub the requests and don't invoke the real services. If this is the case, I would not
be too explicit here. 
If we need these fixtures, could you please explain why? I might be missing something.
though.

spec/features/date_selection_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+          { "date" => (to + 1.day).to_s, "value" => 80 }
+        ]
+      })
+  end
+
+  def expect_metrics_for_each_date_to_be_correct(from, to)
+    month_and_date_string_for_date1 = (from - 1.day).to_s.last(5)
+    month_and_date_string_for_date2 = (from - 2.days).to_s.last(5)
+    month_and_date_string_for_date3 = (to + 1.day).to_s.last(5)
+    unique_pageviews_rows = find("#unique_pageviews_table").all('tr')
```
pmanrubia: Could you please reuse @peteglondon helper for 
getting all the values from a table?

spec/features/date_selection_spec.rb
KyleMacPherson: No, you are correct! I have left in a payload in this example as we need it to be different from the default payload in order to differentiate between the initial payload received when we first load the page and the payload we expect back after we have filter (i think!).

PR #52, opened 2018-09-11
Render the remaining metrics summaries/charts

peteglondon:
Add metric summary/chart for 

* number_of_internal_searches
* satisfaction_score

Also removed the unnecessary `metrics` parameter from the route as we don't need to configure which metrics we are showing and having to specify them is making it really annoying to test.

### DIFF
--- a/app/services/metrics_service.rb
+++ b/app/services/metrics_service.rb
@@ -2,7 +2,8 @@ require 'gds_api/content_data_api'
 
 class MetricsService
   DEFAULT_METRICS = %w[pageviews unique_pageviews number_of_internal_searches satisfaction_score].freeze
-  def fetch(base_path:, from:, to:)
+
+  def fetch_aggregated_data(base_path:, from:, to:)
     url = api.request_url(base_path: base_path, from: from, to: to, metrics: DEFAULT_METRICS)
     api.client.get_json(url).to_hash
   end
--- a/app/services/metrics_service.rb
+++ b/app/services/metrics_service.rb
@@ -1,13 +1,14 @@
 require 'gds_api/content_data_api'
 
 class MetricsService
-  def fetch(base_path:, from:, to:, metrics:)
-    url = api.request_url(base_path: base_path, from: from, to: to, metrics: metrics)
+  DEFAULT_METRICS = %w[pageviews unique_pageviews number_of_internal_searches satisfaction_score].freeze
+  def fetch(base_path:, from:, to:)
+    url = api.request_url(base_path: base_path, from: from, to: to, metrics: DEFAULT_METRICS)
     api.client.get_json(url).to_hash
   end
 
-  def fetch_time_series(base_path:, from:, to:, metrics:)
-    url = api.time_series_request_url(base_path: base_path, from: from, to: to, metrics: metrics)
+  def fetch_time_series(base_path:, from:, to:)
+    url = api.time_series_request_url(base_path: base_path, from: from, to: to, metrics: DEFAULT_METRICS)
     api.client.get_json(url).to_hash
   end
 
--- a/app/views/metrics/show.html.erb
+++ b/app/views/metrics/show.html.erb
@@ -10,13 +10,13 @@
 </div>
 
 <h2>Performance</h2>
-<%= render partial: 'metric_header', locals: {title: 'Unique Pageviews', value: @summary.unique_pageviews } %>
+<%= render 'metric_header', title: 'Unique Pageviews', value: @summary.unique_pageviews  %>
 <%= render_chart @summary.unique_pageviews_series %>
-<%= render partial: 'metric_header', locals: { title: 'Pageviews', value: @summary.pageviews } %>
+<%= render 'metric_header',  title: 'Pageviews', value: @summary.pageviews  %>
 <%= render_chart @summary.pageviews_series %>
-<%= render partial: 'metric_header', locals: { title: 'Searches from the page', value: @summary.number_of_internal_searches } %>
+<%= render 'metric_header',  title: 'Searches from the page', value: @summary.number_of_internal_searches  %>
 <%= render_chart @summary.number_of_internal_searches_series %>
 
 <h2>Feedback</h2>
-<%= render partial: 'metric_header', locals: { title: 'User satisfaction score', value: @summary.satisfaction_score } %>
+<%= render 'metric_header', title: 'User satisfaction score', value: @summary.satisfaction_score  %>
 <%= render_chart @summary.satisfaction_score_series %>
--- a/app/views/metrics/show.html.erb
+++ b/app/views/metrics/show.html.erb
@@ -10,13 +10,21 @@
 </div>
 
 <h2>Performance</h2>
-<%= render 'metric_header', title: 'Unique Pageviews', value: @summary.unique_pageviews  %>
+<div class="metric_summary unique_pageviews">
+  <%= render 'metric_header', title: 'Unique Pageviews', value: @summary.unique_pageviews %>
+</div>
 <%= render_chart @summary.unique_pageviews_series %>
-<%= render 'metric_header',  title: 'Pageviews', value: @summary.pageviews  %>
+<div class="metric_summary pageviews">
+  <%= render 'metric_header', title: 'Pageviews', value: @summary.pageviews %>
+</div>
 <%= render_chart @summary.pageviews_series %>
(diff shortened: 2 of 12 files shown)


peteglondon:
[(3) Show metadata information in the Single Performance Page (EPIC 2)](https://trello.com/c/hjj6IJXH/621-3-show-metadata-information-in-the-single-performance-page-epic-2)

app/views/metrics/show.html.erb
```diff
@@ -2,15 +2,21 @@
 <h2>Page Data</h2>
 <h1><%= @summary.title %></h1>
 <div class="page-metadata">
-  <%= render :partial => 'metadata_row', locals: { label: 'Published', value: @summary.published_at } %>
-  <%= render :partial => 'metadata_row', locals: { label: 'Last updated', value: @summary.last_updated } %>
-  <%= render :partial => 'metadata_row', locals: {label: 'From', value:  @summary.publishing_organisation } %>
-  <%= render :partial => 'metadata_row', locals: {label: 'Type', value:  @summary.document_type } %>
-  <%= render :partial => 'metadata_row', locals: {label: 'URL', value:  @summary.base_path } %>
+  <%= render partial: 'metadata_row', locals: { label: 'Published', value: @summary.published_at } %>
```
pmanrubia: Not sure if we need `partial` and `locals`?

peteglondon: Done

spec/features/single_content_item_spec.rb
```diff
@@ -46,6 +57,18 @@
       expect(page).to have_content('200000')
     end
 
+    it 'renders a metric for satisfaction_score' do
+      expect(page).to have_content('25.5')
```
pmanrubia: This test relies on the existence of `25.5` in the whole page. It is better to use selectors:

```ruby
expect(page).to have_selector('css_selector', text: '25.5')
```

If not, we rely on extending the feature without having duplicated values.

spec/features/single_content_item_spec.rb
```diff
@@ -46,6 +57,18 @@
...
+    end
+
+    it 'renders the page title' do
+      expect(page).to have_content('Content Title')
```
pmanrubia: This test relies on the existence of `Content Title` in the whole page. It is better to use selectors:

```ruby
expect(page).to have_selector('css_selector', text: 'Content Title')
```

If not, we rely on extending the feature without having duplicated values.

spec/features/single_content_item_spec.rb
```diff
@@ -46,6 +57,18 @@
...
+    end
+
+    it 'renders a metric for on page searches' do
+      expect(page).to have_content('250')
```
pmanrubia: This test relies on the existence of `250` in the whole page. It is better to use selectors:

```ruby
expect(page).to have_selector('css_selector', text: '250')
```

If not, we rely on extending the feature without having duplicated values.

spec/features/single_content_item_spec.rb
```diff
@@ -1,12 +1,12 @@
 RSpec.describe '/metrics/base/path', type: :feature do
   include GdsApi::TestHelpers::ContentDataApi
-
+  let(:metrics) { MetricsService::DEFAULT_METRICS }
```
pmanrubia: I would rather have explicit fixtures and be explicit setting up the fixtures (do not rely on production code). 
In this case, we can be specific on line 3 with the metrics that we will be asking for, and then, in the application code we just use a constant.
The reason behind this is: if we decide to change the way we store the metrics (not use a constant anymore), then I don't have to change the test because it would be risky.

app/views/metrics/show.html.erb
```diff
@@ -2,15 +2,21 @@
...
-Unique Pageviews: <%= @summary.unique_pageviews %>
+<h2>Performance</h2>
+<%= render 'metric_header', title: 'Unique Pageviews', value: @summary.unique_pageviews  %>
 <%= render_chart @summary.unique_pageviews_series %>
-
-Pageviews: <%= @summary.pageviews %>
+<%= render 'metric_header',  title: 'Pageviews', value: @summary.pageviews  %>
 <%= render_chart @summary.pageviews_series %>
+<%= render 'metric_header',  title: 'Searches from the page', value: @summary.number_of_internal_searches  %>
+<%= render_chart @summary.number_of_internal_searches_series %>
```
pmanrubia: Not sure why we have a helper method for `render_chart`. If we don't have data, then we should show an empty chart (which is consistent with the existing design)

```ruby
   render "components/chart", series.chart_data if series.has_values?
```
I would simply do:

```ruby
  render "components/chart", series.chart_data 
```

app/services/metrics_service.rb
```diff
@@ -1,13 +1,14 @@
 require 'gds_api/content_data_api'
 
 class MetricsService
-  def fetch(base_path:, from:, to:, metrics:)
-    url = api.request_url(base_path: base_path, from: from, to: to, metrics: metrics)
+  DEFAULT_METRICS = %w[pageviews unique_pageviews number_of_internal_searches satisfaction_score].freeze
+  def fetch(base_path:, from:, to:)
```
pmanrubia: The method `fetch` does not reveal very well what it does. I know this is not something that you have introduced, but this might be a good opportunity to rename it. Effectively it just retrieves AggregatedMetrics, so it would rename it to better represent what it does.

This might be a good opportunity to split this Service into two:

1. FindTimeSeries
2. FindAggregatedValues

spec/features/single_content_item_spec.rb
```diff
@@ -66,28 +89,54 @@
...
       expect(pageviews_rows[2].text).to eq '01-14 20'
       expect(pageviews_rows[3].text).to eq '01-15 30'
     end
+
+    it 'renders the metric timeseries for on-page searches' do
+      click_on 'Number of internal searches table'
+      internal_searches_rows = find("#number_of_internal_searches_2018-01-13-2018-01-15_table").all('tr')
+
+      expect(internal_searches_rows.count).to eq 4
+      expect(internal_searches_rows[0].text).to eq ''
```
pmanrubia: Have you thought of using the `Rspec` matcher `match_array`.
We would dry it up a lot by using:

```ruby
  expect(internal_searches_rows).to match_array([-----])
```

spec/features/single_content_item_spec.rb
peteglondon: Done

spec/features/single_content_item_spec.rb
peteglondon: Done

spec/features/single_content_item_spec.rb
peteglondon: Refactored to use array matcher as follows

```ruby
expect(unique_pageviews_rows).to match_array([
        ['', ''],
        ['01-13', '101'],
        ['01-14', '202'],
        ['01-15', '303']
      ])
```

app/services/metrics_service.rb
peteglondon: I've renamed the method to `fetch_aggregated_data`.

The only reason i can see to split this class into 2 at the moment apart from it has 2 methods.

I'm not sure this is a good enough reason to split it.

spec/features/single_content_item_spec.rb
```diff
@@ -65,20 +84,57 @@
...
+        ['', ''],
+        ['01-13', '10'],
+        ['01-14', '20'],
+        ['01-15', '30']
+      ])
+    end
+
+    it 'renders the metric timeseries for on-page searches' do
+      click_on 'Number of internal searches table'
+      internal_searches_rows = find("#number_of_internal_searches_2018-01-13-2018-01-15_table").all('tr').map do |el|
```
pmanrubia: Can you extract this into a support method that returns the table content, this way we can reuse it anywhere.

As a side goal, can you please simplify the CSS selector? You might need to wrap the chart with a div, or add other classes to the table, but `number_of_internal_searches_2018-01-13-2018-01-15_table` feels like too complex for a test CSS selector.

app/views/metrics/show.html.erb
peteglondon: I've had a word with FE dev and Designer this end. I seems showing an empty chart (and an empty table) when we have no data is not desirable. 
The preferred option is to show a div with a message to indicate there is no data (to be styled later).

spec/features/single_content_item_spec.rb
peteglondon: Done

PR #56, opened 2018-09-13
Hook up components

peteglondon:
This commit shows the data in the components we have so far.

The components we are now using:

* components/metadata
* components/info-metric
* components/chart

## Before
before

## After
Stuff in green - real data
Stuff in red - placeholders
after

Url: http://content-data-admin.dev.gov.uk/metrics/check-mot-history?from=2018-01-01&to=2018-10-01

### DIFF
--- a/app/views/metrics/_metric_header.html.erb
+++ b/app/views/metrics/_metric_header.html.erb
@@ -1,4 +1,12 @@
-<h3><%= title %></h3>
-<div class="aggregated-value">
-  <%= value %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render "components/info-metric", {
+        name: title,
+        figure: number_with_delimiter(value, delimiter: ','),
+        context: "Number of sessions during which the page was viewed at least once",
+        trend_percentage: 0,
+        period: "From previous 30 days",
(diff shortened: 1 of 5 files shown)


peteglondon:
[(5) Add metrics Charts for the Single Performance Page (EPIC 2)](https://trello.com/c/xZb8cKyy/625-5-add-metrics-charts-for-the-single-performance-page-epic-2)

app/views/metrics/_metric_header.html.erb
```diff
@@ -1,4 +1,12 @@
-<h3><%= title %></h3>
-<div class="aggregated-value">
-  <%= value %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render "components/info-metric", {
+        name: title,
+        figure: number_with_delimiter(value, :delimiter => ','),
```
erino: What about using 1.9 hash syntax here?

chao-xian: @peteglondon can you address Erin's comment please and then we can merge. Not sure why the linter didn't spot that. Thank @erino

erino: Rubocop doesn't check erb, so there maybe other style issues.

peteglondon: Ok I didn't spot that, done now.

PR #59, opened 2018-09-18
Add outline of index page at /content

peteglondon:
We now have a route that displays content items.

Example Url:

/content?date_range=last-30-days&organisation=6667cce2-e809-4e21-ae09-cb0bdc1ddda

This is not complete, This is an interim PR to allow the other stories on the index page 
to be played.

There is a [Related PR iin content-performance-manager](https://github.com/alphagov/content-performance-manager/pull/919)

### DIFF
--- a/app/controllers/content_controller.rb
+++ b/app/controllers/content_controller.rb
@@ -1,0 +1,21 @@
+class ContentController < ApplicationController
+  def index
+    @content = get_content
+  end
+
+private
+
+  def get_content
+    response = metrics_service.content_summary(date_range: date_range,
+      organisation: params[:organisation])
+    @content = ContentItemsPresenter.new(response.deep_symbolize_keys[:results])
+  end
+
+  def metrics_service
+    @metrics_service ||= MetricsService.new
+  end
+
+  def date_range
+    DateRange.new(params[:date_range])
+  end
+end
--- a/app/presenters/content_row_presenter.rb
+++ b/app/presenters/content_row_presenter.rb
@@ -1,0 +1,19 @@
+class ContentRowPresenter
+  attr_reader :title, :base_path, :document_type, :unique_pageviews,
+              :user_satisfaction_score, :number_of_internal_searches
+  def initialize(data)
+    @title = data[:title]
+    @base_path = data[:base_path]
+    @document_type = data[:document_type].try(:tr, '_', ' ').try(:capitalize)
+    @unique_pageviews = data[:unique_pageviews]
+    @user_satisfaction_score = format_satisfaction_score(data[:satisfaction_score], data[:satisfaction_score_responses])
+    @number_of_internal_searches = data[:number_of_internal_searches]
+  end
+
+private
+
+  def format_satisfaction_score(score, responses)
+    return 'No responses' unless score
+    "#{(score * 100).round(1)}% (#{responses} responses)"
+  end
+end
--- a/app/services/metrics_service.rb
+++ b/app/services/metrics_service.rb
@@ -13,9 +13,19 @@ class MetricsService
     api.client.get_json(url).to_hash
   end
 
+  def content_summary(date_range:, organisation:)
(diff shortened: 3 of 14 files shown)


peteglondon:
[(5) Index page: show 100 first content items (EPIC 5)](https://trello.com/c/HvQXxa4z/659-5-index-page-show-100-first-content-items-epic-5)

app/controllers/content_controller.rb
```diff
@@ -0,0 +1,21 @@
+class ContentController < ApplicationController
+  def index
+    @content = get_content
+  end
+
+private
+
+  def get_content
+    response = metrics_service.content_summary(date_range: date_range,
```
pmanrubia: ```MetricsServic``` is doing to much. It might be better to move
into ServiceObjects that do only one thing.

I would be keen to be able to rewrite the ```index``` action into:

```ruby
def action
  response = FindContent.call(params)
  @content = ContentItemsPresenter.new(response)
end
```

```FindContent``` service would live in ```/services/find_content.rb```, and would 
parse the parameters

File: app/controllers/content_controller.rb [hunk 2]
```diff
@@ -0,0 +1,21 @@
...
+      organisation: params[:organisation])
+    @content = ContentItemsPresenter.new(response.deep_symbolize_keys[:results])
+  end
+
+  def metrics_service
```
pmanrubia: I would break the ```MetricsService``` into services for the other two methods:
1. fetch series
2. fetch aggregations

app/controllers/content_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    @metrics_service ||= MetricsService.new
+  end
+
+  def date_range
```
pmanrubia: This code could also live within the service.

app/presenters/content_row_presenter.rb
```diff
@@ -0,0 +1,19 @@
+class ContentRowPresenter
```
pmanrubia: Could you please add a unit test for this class?
It is good to have a descriptive test to describe code like:

```ruby
  @document_type = data[:document_type].try(:tr, '_', ' ').try(:capitalize)
```

app/services/metrics_service.rb
```diff
@@ -13,9 +13,19 @@ def fetch_time_series(base_path:, date_range:)
...
+    api_with_long_timeout.content_summary(from: date_range.from, to: date_range.to, organisation: organisation)
+  end
+
 private
 
   def api
     @api ||= GdsApi::ContentDataApi.new
   end
+
+  def api_with_long_timeout
```
pmanrubia: Is there any reason for not having a long time-out for
all our requests?. It makes sense to me to follow the 
same approach for all our requests.

app/services/metrics_service.rb
```diff
@@ -13,9 +13,19 @@ def fetch_time_series(base_path:, date_range:)
...
+    api.tap do |client|
+      client.options[:timeout] = 30
```
pmanrubia: Not sure if `30 seconds` of timeout on the client side
matches the timeout defined in Nginx.

lib/gds_api/content_data_api.rb
```diff
@@ -19,12 +19,18 @@ def content_data_api_endpoint
     "#{Plek.current.find('content-performance-manager')}/api/v1"
   end
 
-  def query(from:, to:, metrics:)
-    query_params = {}
-    query_params[:from] = from
-    query_params[:to] = to
-    query_params[:metrics] = metrics
+  def content_summary(from:, to:, organisation:)
```
pmanrubia: Not sure if `content_summary` represents what the endpoint does.
Can you think of a better name? It doesn't feel to me like a summary,
but a list of content items.

app/services/metrics_service.rb
peteglondon: No that should have been removed

app/services/metrics_service.rb
peteglondon: Again this is redundant.

PR #60, opened 2018-09-18
Fix format of percentage numbers

KyleMacPherson:
The percentage figures for metrics are not presented as percentages but as whole numbers on the single page view.
This PR fixes that.

https://trello.com/c/DhJHJL6b/679-1-fix-formatting-of-numbers-percentages-for-single-page

### DIFF
--- a/app/presenters/chart_presenter.rb
+++ b/app/presenters/chart_presenter.rb
@@ -44,6 +44,8 @@ class ChartPresenter
 
   def values
     return [] unless json[metric]
-    json[metric].map { |hash| hash['value'] }
+    json[metric].map do |hash|
+      metric == :satisfaction_score ? (hash['value'] * 100).round : hash['value']
+    end
   end
 end
--- a/app/presenters/chart_presenter.rb
+++ b/app/presenters/chart_presenter.rb
@@ -1,4 +1,5 @@
 class ChartPresenter
+  include MetricsFormatterHelper
   attr_reader :json, :metric, :from, :to
 
   def initialize(json:, metric:, from:, to:)
@@ -44,8 +45,6 @@ class ChartPresenter
 
   def values
     return [] unless json[metric]
-    json[metric].map do |hash|
-      metric == :satisfaction_score ? (hash['value'] * 100).round : hash['value']
-    end
+    json[metric].map { |hash| format_metric_value(metric, hash['value']) }
   end
 end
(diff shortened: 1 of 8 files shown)


KyleMacPherson:
[(1) Fix formatting of numbers/percentages (for single page)](https://trello.com/c/DhJHJL6b/679-1-fix-formatting-of-numbers-percentages-for-single-page)

app/presenters/chart_presenter.rb
```diff
@@ -42,6 +42,8 @@ def keys
   end
 
   def values
-    json[metric].map { |hash| hash['value'] }
+    json[metric].map do |hash|
+      metric == :satisfaction_score ? (hash['value'] * 100).round : hash['value']
```
pmanrubia: For the time being can we extract a helper to render the satisfaction score?

```ruby
json[metric].map { |hash| MetricsFormatterHelper.format(metric, hash['value']) }
```

We should be able to use it in other places as well.

app/presenters/single_content_item_presenter.rb
```diff
@@ -19,7 +19,7 @@ def parse_metrics(metrics)
     @unique_pageviews = metrics[:unique_pageviews]
     @pageviews = metrics[:pageviews]
     @number_of_internal_searches = metrics[:number_of_internal_searches]
-    @satisfaction_score = metrics[:satisfaction_score]
+    @satisfaction_score = 100 * metrics[:satisfaction_score]
```
pmanrubia: I would rely on the helper. See comment above.

app/views/components/_info-metric.html.erb
```diff
@@ -7,11 +7,12 @@
   context ||= false
 
 %>
+<% percentage_metrics = ['User satisfaction score'] %>
 <% if name && figure && context %>
   <div class="app-c-info-metric govuk-body">
     <h3 class="app-c-info-metric__heading"><%= name %></h3>
     <p class="app-c-info-metric__context govuk-body-s"><%= context %></p>
-    <span class="app-c-info-metric__figure govuk-!-font-weight-bold govuk-!-font-size-36 "><%= figure %></span>
+    <span class="app-c-info-metric__figure govuk-!-font-weight-bold govuk-!-font-size-36 "><%= figure %><%= '%' if percentage_metrics.include? name %></span>
```
pmanrubia: I am not clear on why we need this. It doesn't feel totally right.

app/presenters/chart_presenter.rb
pmanrubia: Also the helper can rely on Rails helper to render numbers / percentages.

PR #61, opened 2018-09-19
Add time select component

maxgds:
This adds the component that allows users to select the timeframe for the data they are viewing.

It also updates the view to use the component, and updates related tests.

This can be seen at 
http://content-data-admin.dev.gov.uk/component-guide/time-select
or single content item pages such as
http://content-data-admin.dev.gov.uk/metrics/government/publications/self-assessment-tax-return-sa100?utf8=%E2%9C%93&date_range=last-2-years
(note that this branch has inherited the satisfaction score bug from master, so you'll need to work around that to view it - remove the satisfaction graph in the view and line 4 of app/services/metrics_service.rb) 

screen shot 2018-09-19 at 09 39 37

### DIFF
--- a/app/views/components/_chart.html.erb
+++ b/app/views/components/_chart.html.erb
@@ -21,12 +21,14 @@
         1 => {lineWidth: 2, lineDashStyle: [10, 2], color: '#454a4c'}
     }
   }
-  chart_format_data = rows.each_with_object([]) do |row, acc|
-    acc << {
-    name: row[:label],
-    linewidth: 10,
-    data: keys.zip(row[:values])
-    }
+  if rows.length > 0 && keys.length > 0
+    chart_format_data = rows.each_with_object([]) do |row, acc|
+      acc << {
+      name: row[:label],
+      linewidth: 10,
+      data: keys.zip(row[:values])
+      }
+    end
   end
 %>
 <% if rows.length > 0 && keys.length > 0 && chart_id && table_id %>
--- a/app/views/components/_time-select.html.erb
+++ b/app/views/components/_time-select.html.erb
@@ -1,0 +1,29 @@
+<%
+  base_path ||= false
+  current_selection ||= false
+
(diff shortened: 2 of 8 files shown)


app/views/components/_time-select.html.erb
```diff
@@ -0,0 +1,29 @@
+<%
+  base_path ||= false
+  current_selection ||= false
+
+  if dates.length > 1
```
KyleMacPherson: [Minor] Was thinking we could put this on line 15 to avoid asserting that `dates.length > 1` twice. What do you think?

app/views/components/_chart.html.erb
```diff
@@ -21,13 +21,14 @@
...
     }
   }
-  chart_format_data = rows.each_with_object([]) do |row, acc|
-    acc << {
-    name: row[:label],
-    linewidth: 10,
-    data: keys.zip(row[:values])
-    }
-  end
+  rows.length > 0 && keys.length > 0
```
KyleMacPherson: I think we might be missing an `if` here?

app/views/components/_time-select.html.erb
maxgds: Not sure. The reason I haven't is that the top section is about getting the data clean and ready to display, the bottom section is about displaying with logic only where necessary for guarding against breakage and allowing decision paths - making the content bold feels like the former, and the second assertion is a guard. Do you think it worth breaking that guideline for the clarity?

app/views/components/_chart.html.erb
maxgds: yep, will fix

app/views/components/_time-select.html.erb
KyleMacPherson: No, that sounds good to me!

maxgds:
@KyleMacPherson fixed the if, good to merge now?

PR #67, opened 2018-09-20
Increase the timeout for GdsApi::ContentDataApi

peteglondon:
We are currently experiencing performance problems
when retrieving data for the index page.

This commit (temporarily) increases the timeout
so we can render the page and development on it
can continue.

We will reduce the timeout again once the
query has been optimised in the content-data-api.

### DIFF
--- a/app/services/metrics_common.rb
+++ b/app/services/metrics_common.rb
@@ -4,7 +4,9 @@ module MetricsCommon
 private
 
   def api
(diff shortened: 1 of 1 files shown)


peteglondon:
[(5) Index page: show 100 first content items (EPIC 5)](https://trello.com/c/HvQXxa4z/659-5-index-page-show-100-first-content-items-epic-5)

PR #69, opened 2018-09-20
Add feedback comments metric to single page view

theseanything:
Requests now support the feedex_comments metric to display the aggregated and time series data for number of feedback comments. Graph and metric info components added to single page view for feedback comments.


app/presenters/chart_presenter.rb
```diff
@@ -39,12 +39,14 @@ def chart_data
 
   def keys
     return [] unless json[metric]
+
```
pmanrubia: 👍 these empty lines make a difference :)

PR #75, opened 2018-09-21
link from index page to single content item page

KyleMacPherson:


pmanrubia:
@KyleMacPherson could you please add a description to the PR in a similar way to: https://github.com/alphagov/content-data-admin/pull/76

spec/features/index_page_spec.rb
```diff
@@ -1,6 +1,7 @@
 RSpec.describe '/content' do
   include GdsApi::TestHelpers::ContentDataApi
   include TableDataSpecHelpers
+  let(:metrics) { %w[pageviews unique_pageviews number_of_internal_searches feedex_comments] }
```
pmanrubia: `metrics` is only use within the test. Would you mind moving it there?
I am also curious about why we need to specify them at this point. Shouldn't we have a default?

spec/features/index_page_spec.rb
```diff
@@ -46,4 +47,11 @@
       ]
     )
   end
+
+  it 'takes you to single content item if you click on title of item' do
+    content_data_api_has_metric(base_path: 'path/1', from: (Time.zone.today - 30.days), to: Time.zone.today, metrics: metrics)
```
pmanrubia: In theory, the `from` and `to` should be the same as the ones specifies in the index page. Are we testing it?

app/views/content/_content_row.html.erb
```diff
@@ -1,5 +1,5 @@
 <tr>
-  <td><%= item.title %>
+  <td><%= link_to item.title, "/metrics#{item.base_path}" %>
```
pmanrubia: If you add the `:as` clause to `routes.rb`:

```ruby
  get '/metrics/*base_path', to: 'metrics#show', as: :metrics
```

You can use the helper for the route:

```ruby
  metrics_path('the/base/path')
```

app/presenters/content_row_presenter.rb
```diff
@@ -17,4 +17,10 @@ def format_satisfaction_score(score, responses)
 
     "#{(score * 100).round(1)}% (#{responses} responses)"
   end
+
+  def format_base_path(base_path)
+    #  remove '/' to make base_path usable in links
```
pmanrubia: Minor: what about using `base_path.delete_prefix('/')`?

PR #76, opened 2018-09-21
Show publishing app in metadata section for the Single Performance Page

pmanrubia:
[Trello card](https://trello.com/c/061BqCh7/688-1-display-the-application-that-published-the-page)

**As a** Content Designer
**I want** to know what application has created the document
**So** that I can edit it with the right application

## What

Use the field `publishing_app` to display the application that edited 
the document. 

In future stories we will provide a link to open the document directly
in the publishing-app, but to keep the story simple, we will just render
the name of the application.

## Why

Adding a link to the publishing app depends on the specific format
of each application, so we would rather take this approach in smaller
steps and prioritise this work accordingly.

### After:

image


KyleMacPherson:
@pmanrubia looks good, just needs a rebase!

PR #77, opened 2018-09-21
Use locales in single page view

theseanything:
This is an initial PR to use locales in the single page view. Locales have not used been *everywhere* as waiting for refactoring template partials.

### DIFF
--- a/config/locales/views/metrics/en.yml
+++ b/config/locales/views/metrics/en.yml
@@ -1,13 +1,14 @@
 en:
-  views:
-    metrics:
-      show:
-        page_kicker: 'Page data'
-        navigation:
-          back_link: 'View all pages'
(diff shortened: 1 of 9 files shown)


config/locales/views/metrics/en.yml
```diff
@@ -1,13 +1,25 @@
...
+      download:
+        download_data: 'Download the data'
+        csv_link: '%{metric_name} CSV'
+      time_periods:
+        last_30_days: 'Past 30 days'
+        last_month: 'Last month'
+        last_3_months: 'Past 3 months'
+        last_6_months: 'Past 6 months'
+        last_year: 'Past year'
+        last_2_years: 'Past 2 years'
```
pmanrubia: End of line

PR #88, opened 2018-09-26
Add metric section partial

theseanything:
Adding a metric section partial to reduce code repetition in the single page view and enable locale use in graph metric overviews.


theseanything:
[(2) Use Rails Translations for interface microcopy](https://trello.com/c/jFr8BQaS/681-2-use-rails-translations-for-interface-microcopy)

pmanrubia:
👍

PR #89, opened 2018-09-26
Use locales in about this data section

theseanything:
Textual infomation around about this data is now render using locales.


theseanything:
[(2) Use Rails Translations for interface microcopy](https://trello.com/c/jFr8BQaS/681-2-use-rails-translations-for-interface-microcopy)

PR #90, opened 2018-09-26
Change link text as per designs

KyleMacPherson:
Remove the word `visit` from the link to content item on gov.uk link on
the single content item page.

The word `visit` is necessary for accessibility purposes but not helpful for
the design so this PR hides it using the visually-hidden class to allow
screen readers to understand it.

Before: 
screen shot 2018-09-26 at 12 27 28
After:
screen shot 2018-09-26 at 12 27 17

### DIFF
--- a/app/views/metrics/show.html.erb
+++ b/app/views/metrics/show.html.erb
@@ -21,12 +21,13 @@
                   {
                     header: t(".navigation.visit_page"),
                     link_url: "//www.gov.uk#{@performance_data.metadata[:base_path]}",
-                    link_text: t('.navigation.visit_link', content_title: @performance_data.title)
+                    accessibility_message: t('.navigation.accessibility_message'),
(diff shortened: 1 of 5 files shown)


app/views/metrics/show.html.erb
```diff
@@ -21,12 +21,14 @@
...
-                    link_text: t('.navigation.visit_link', content_title: @performance_data.title)
+                    accessibility_message: t('.navigation.accessibility_message'),
+                    label: @performance_data.title
                   },
                   {
                     header: "Make changes to the page",
                     link_url: "#",
-                    link_text: @performance_data.publishing_app
+                    accessibility_message: "",
+                    label: @performance_data.publishing_app
```
theseanything: Should we use the locales for the edit link lable? In the design it shows with `Edit <publisher app>`

PR #114, opened 2018-10-09
Add on page search rate

KyleMacPherson:
### What does this PR do?

Calculates the on page search rate. This is used in _'X% of users searched on this page'_. (See screenshot).

### Why?

To give users some context around the search statistics to help them understand the relevance.

### Technical Note

I considered making changes to the way we handle `@context_metrics` within `GlanceMetricPresenter.rb`.

I'm not sure that `@context_metrics` needs to be a hash, maybe we can simply render `@context_metrics` as a string depending on what the metric name is. In the future I think this will avoid running multiple unnecessary methods.

If the rest of the team thinks this is a good idea then I will address this in a separate PR.

### How it looks
screen shot 2018-10-09 at 10 33 46

### DIFF
--- a/app/presenters/glance_metric_presenter.rb
+++ b/app/presenters/glance_metric_presenter.rb
@@ -1,15 +1,16 @@
 class GlanceMetricPresenter
-  def initialize(metric_name, metric_value, time_period)
+  def initialize(metric_name, metric_value, time_period, secondary_metric_value = nil)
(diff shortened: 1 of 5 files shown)


KyleMacPherson:
[(2) Page data - Display how many users searched the page](https://trello.com/c/KyggchDS/692-2-page-data-display-how-many-users-searched-the-page)

PR #117, opened 2018-10-10
Use single endpoint

theseanything:
Updating the single page presenter to use the new `/single_page` endpoint from content performance manager. Removed old request test stubs for access in the previous endpoints.


theseanything:
[(5) Page data - Use a Single endpoint to return data](https://trello.com/c/X4TC7itU/689-5-page-data-use-a-single-endpoint-to-return-data)

PR #124, opened 2018-10-16
Allow organisation filtering to respect date range

KyleMacPherson:
### Why?
[Trello card](https://trello.com/c/nQxldbA2/661-1-index-page-filter-by-organisation)
Currently after filtering if the user clicks on a content item title the page will
not load due to an empty date range param. This bug was discovered after we added
the filter by organisation functionality.

### What?
This PR allows for empty date range params and persists any existing date range params
i.e if in the future we filter the index page to show data from the last 2 years then the page data
page will reflect this.

### DIFF
--- a/app/controllers/content_controller.rb
+++ b/app/controllers/content_controller.rb
@@ -13,6 +13,7 @@ private
   end
 
   def date_range
-    DateRange.new(params[:date_range])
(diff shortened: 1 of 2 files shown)


KyleMacPherson:
[(1) Index page - Filter by organisation](https://trello.com/c/nQxldbA2/661-1-index-page-filter-by-organisation)

PR #125, opened 2018-10-16
Update time select component to allow it to be embedded in another form.

maxgds:
https://trello.com/c/MC1QwkUL/738-3-index-page-update-date-select-component-so-it-becomes-part-of-another-form

We want to use this form section within the form on the index page, since it was a standalone form originally this wouldn't have been possible so in this PR we move the form tags into the view template leaving the component as a form partial. In the original view the component also rendered the submit button and because it is hidden in the reveal, I have had to preserve that ability. Since we don't want it on the index page, though, it is now parameterised. Visually there are no changes.

screen shot 2018-10-16 at 14 13 08

### DIFF
--- a/app/views/components/_time-select.html.erb
+++ b/app/views/components/_time-select.html.erb
@@ -1,6 +1,7 @@
 <%
   base_path ||= false
   current_selection ||= false
+  render_button ||= false
   submit_button_text = t ".submit_button"
 
   if dates.length > 1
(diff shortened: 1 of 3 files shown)


app/views/components/_time-select.html.erb
```diff
@@ -12,19 +13,20 @@
...
+
         <%= render "govuk_publishing_components/components/radio", {
           name: "date_range",
           items: dates
         } %>
-        <%= render "govuk_publishing_components/components/button", {text: submit_button_text} %>
-      <% end %>
+        <% if render_button != false %>
+          <%= render "govuk_publishing_components/components/button", {text: submit_button_text} %>
+        <% end %>
```
theseanything: Should we just add the button to the form of the in `show.html.erb` rather than adding `render_button` - as the button only makes sense in when embedded in the form.

maxgds: We can't because it has to appear in the reveal and not outside of it - else you get the header "Last 30 days" and a submit button without seeing the rest of the form.

theseanything: Ah sure!

PR #127, opened 2018-10-17
Add styling for mobile views, tweak desktop view to match design.

maxgds:
https://trello.com/c/csZsWZNy/736-3-page-data-make-layout-work-at-mobile-and-tablet-breakpoints

Matching styles against designs for desktop and mobile:

https://drive.google.com/open?id=194icCuUprUsYYtakbUHGZbDQySoLnCSe
https://drive.google.com/open?id=1IHYvkjQm_ZAU9Er-cIdAapNLZ7qEDOns

Some changes to CSS classes, building on top of recent changes in the gem.

### DIFF
--- a/app/assets/stylesheets/components/_glance-metric.scss
+++ b/app/assets/stylesheets/components/_glance-metric.scss
@@ -13,6 +13,8 @@
 .app-c-glance-metric__heading {
   @include govuk-responsive-margin(3, "top");
   @include govuk-responsive-margin(2, "bottom");
+  min-height: 2.6em;
+
 
   @include media-down(tablet) {
     min-height: 4em;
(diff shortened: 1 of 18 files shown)


maxgds:
@alex-ju I think this is ready again for review.

app/assets/stylesheets/components/_glance-metric.scss
```diff
@@ -13,6 +13,8 @@
 .app-c-glance-metric__heading {
   @include govuk-responsive-margin(3, "top");
   @include govuk-responsive-margin(2, "bottom");
+  min-height: 2.6em;
+
 
   @include media-down(tablet) {
```
alex-ju: Non-blocker: instead of `media-down(tablet)` — which is a mixin that we will deprecate — you can use the govuk-frontend version of it, which is `@include govuk-media-query($from: tablet)`; `mobile`, `tablet` and `desktop`) being the breaking points, always going up as the intended approach is to go mobile first. This need quite a bit of refactoring, so it's not a blocking comment, but more for awareness.

package.json
```diff
@@ -1,5 +1,7 @@
 {
   "name": "content-data-admin",
   "private": true,
-  "dependencies": {}
+  "dependencies": {
+    "govuk-frontend": "^2.2.0"
```
alex-ju: we don't need this bit, govuk-frontend is available through govuk_publishing_components. in fact I think you don't need this file at all

app/assets/stylesheets/components/_glance-metric.scss
maxgds: Fair enough - this stemmed partially from us having no mobile designs when I built the first iteration. I'll bear it in mind for future and switch out as I go when revisiting the CSS, just so we can push on for now.

PR #130, opened 2018-10-18
Show trend metrics

theseanything:
This PR enables the single page view to show the trend percentages for metrics using data supplied from the content-performance-manager. Changes include:
- Refactoring DateRange class to return previous time period
- Adding another API call to content-performance-manager for previous period
- Adding methods to single page presenter to calculate trend metrics
- Add feature tests for trend metrics
- Refactor API stubs in testing to account for extra API call
- Fix issue where stubs return response with dates outside date range


theseanything:
[(5) Page data - Show stats compared to previous period](https://trello.com/c/6ZhsvKml/710-5-page-data-show-stats-compared-to-previous-period)

PR #131, opened 2018-10-18
Fix single page layout

theseanything:
This fixes the following issues on the single page view:
- Section order of the metrics 
- Section headings
- The line width of the `about` metric descriptions


app/views/metrics/_metric_header.html.erb
```diff
@@ -22,5 +22,8 @@
     options.update({ short_context: short_context })
   end
 %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
 <%= render "components/info-metric", options %>
```
maxgds: please just tweak the indent here and push the render in a little

PR #132, opened 2018-10-18
Improve make changes to the page

cbaines:
Make the URL functional, at least for Whitehall Publisher, and
Publisher content. For other publishing apps, it won't appear, and
should be extended when possible.

To help with isolating the logic, I've moved it in to a helper. This
makes the choice regarding what URL and label to provide to the view.

---

example1
example2


app/views/metrics/show.html.erb
```diff
@@ -25,18 +25,21 @@
...
+            accessibility_message: t('.navigation.accessibility_message'),
+            label: @performance_data.title
+          },
+          {
+            header: "Make changes to the page",
+            link_url: edit_url_for(
+                        content_id: @performance_data.content_id,
+                        publishing_app: @performance_data.publishing_app
+                      ),
+            label: edit_label_for(publishing_app: @performance_data.publishing_app)
```
theseanything: Can `edit_url_for` and `edit_label_for` be called in the presenter?

app/helpers/edit_content_helper.rb
```diff
@@ -0,0 +1,24 @@
+module EditContentHelper
```
theseanything: I feel this module name is a bit ambiguous/confusing, maybe something like `ExternalLinksHelper`?

config/locales/defaults/en.yml
```diff
@@ -4,6 +4,10 @@ en:
     calculated_google_analytics: 'calculated from Google Analytics'
     feedback_explorer: 'Feedback Explorer'
     none: false
+  edit_content:
+    edit_page: 'Make changes to the page'
+    edit_link: 'Edit in %{publishing_app}'
+    publisher_edit_link: 'Request a change to mainstream content'
```
theseanything: I feel content designers will expect these to be in `locales/views/metrics/en.yaml`

app/views/metrics/show.html.erb
cbaines: I guess with the limited Rails experience I have, I'm more used to using helper modules in controllers or views, but looking at the only other helper (MetricsFormatterHelper), it's used in presenters.

I'm still not quite sure how I feel about `SingleContentItemPresenter`, and what it's responsibility is, but I have added methods to call the functions in the helper module now.

app/helpers/edit_content_helper.rb
cbaines: I'm ok with that, and I've changed the naming.

config/locales/defaults/en.yml
cbaines: I moved them out of the metrics specific file when I moved the code to the helper, but I'm fine with keeping them in the metrics related file as well. I've moved them back now.

app/views/metrics/show.html.erb
```diff
@@ -25,18 +25,18 @@
...
+          {
+            header: t(".navigation.visit_page"),
+            link_url: "//www.gov.uk#{@performance_data.base_path}",
+            accessibility_message: t('.navigation.accessibility_message'),
+            label: @performance_data.title
+          },
+          {
+            header: t(".navigation.edit_page"),
+            link_url: @performance_data.edit_url,
+            label: @performance_data.edit_label
```
theseanything: Should we add the accessibilty message?

spec/presenters/single_content_item_presenter_spec.rb
```diff
@@ -10,18 +10,6 @@
...
-    end
-
-    it 'capitalizes the publishing_app if present' do
-      expect(subject.publishing_app).to eq('Publisher')
-    end
-  end
-
   describe '#trend_percentage' do
     it 'calculates an increase percentage change' do
       current_period_data[:time_series_metrics] = [{ name: 'upviews', total: 100 }]
```
theseanything: Can we add tests for edit_url and edit_label?

cbaines: Sure, I've added a couple of basic tests now. Let me know what you think https://github.com/alphagov/content-data-admin/pull/132/files#diff-2495b53354dceef543a1931fbc51b5ceR71

app/views/metrics/show.html.erb
cbaines: I hadn't looked at that, but I've now had a look.

From what I can tell, this adds some useful context to the links to GOV.UK content (as otherwise it's just the title of the content), but for this link the text will either be "Request a change to mainstream content" or "Edit in #{app}", which seem OK without any additional clarification.

cbaines:
This depends on https://github.com/alphagov/content-performance-manager/pull/965 so we should probably hold off merging until that's merged.

PR #137, opened 2018-10-22
Feedback explorer link

theseanything:
Adding a link to the feedback explorer in the feedback comments section. 
image

Note: the spacing is wider than the design due to complexity issues of custom spacing. This was agreed as and should be revisted in a future PR.

### DIFF
--- a/app/presenters/single_content_item_presenter.rb
+++ b/app/presenters/single_content_item_presenter.rb
@@ -65,6 +65,19 @@ class SingleContentItemPresenter
     I18n.t("metrics.feedex.context")
   end
 
+  def feedback_explorer_href
+    host = Plek.new.external_url_for('support')
+    template = Addressable::Template.new(
+      "#{host}/anonymous_feedback{?from,to,paths}"
+    )
+    query = {
+      paths: base_path,
+      from: @date_range.from,
+      to: @date_range.to
+    }
+    template.expand(query).to_s
+  end
+
   def period
     I18n.t("metrics.show.time_periods.#{@date_range.time_period}.reference")
   end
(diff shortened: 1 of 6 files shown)


app/presenters/single_content_item_presenter.rb
```diff
@@ -65,6 +65,18 @@ def feedex_context
     I18n.t("metrics.feedex.context")
   end
 
+  def feedback_explorer_href
+    template = Addressable::Template.new(
+      "#{Plek.find('support')}/anonymous_feedback{?from,to,paths}"
```
cbaines: I think here, `Plek.new.external_url_for`is the thing to use, as that will generate the right URL for the Support app in AWS environments which use different URLs for internal and external traffic.

app/views/metrics/_metric_section.html.erb
```diff
@@ -8,4 +8,14 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-one-half">
+    <a href="" class="govuk-link govuk-body-s"><%= t ".download_link", metric_name: t("metrics.#{metric_name}.title") %></a>
+  </div>
+  <div class="govuk-grid-column-one-half">
+    <% if external_link %>
+      <a href="<%= external_link %>" class="govuk-link govuk-body-s"><%= t "metrics.#{metric_name}.external_link" %></a>
+    <% end %>
+  </div>
+</div>
```
cbaines: Missing trailing newline here.

app/presenters/single_content_item_presenter.rb
theseanything: Yup! Changed to `external_url_for`

PR #138, opened 2018-10-22
Correct capitalisation on dates and heading

pollygreen:


theseanything:
I assume this was intending to correct casing for the heading circled in red. But need check that is doesn't effect headings circled in green.
time-select

theseanything:
re-raise in #144

PR #142, opened 2018-10-23
Rename table reveal

pollygreen:
Removed 'View' from the beginning of the table reveal description to reflect the design.


theseanything:
re-raised in #146

PR #151, opened 2018-10-24
Fix search rate bug

KyleMacPherson:
### What?

[Trello card](https://trello.com/c/TdGevWTq/764-1-page-data-fix-of-users-searched-the-page-in-the-at-a-glance-metrics)

This PR is to fix a bug (see trello card) which has crept into the code base resulting in `search_rate` for a content item being calculated incorrectly.

Search rates were being calculated incorrectly due to our use of the method
`number_with_delimiter` which, for example, returns '10,000' when passed 10000.

Calling `to_f` on '10,000' returns 10 and this is the cause of our current incorrect
calculation for search_rate on some pages.

### How?

This is solved by formatting the `total_searches` and `total_pviews` into
a float before using them to calculate `on_page_search_rate`.

This PR adds a test for this functionality and makes changes to default data in order
to test that this bug is not longer a problem.

### DIFF
--- a/app/presenters/single_content_item_presenter.rb
+++ b/app/presenters/single_content_item_presenter.rb
@@ -146,11 +146,11 @@ private
   end
 
   def on_page_search_rate
-    metric_value = self.total_searches
-    secondary_metric_value = self.total_pviews
+    metric_value_as_float = self.total_searches.delete(',').to_f
+    secondary_metric_value_as_float = self.total_pviews.delete(',').to_f
 
-    return 0 if metric_value.to_i.zero? || secondary_metric_value.to_i.zero?
-    search_rate = (metric_value.to_f / secondary_metric_value.to_f) * 100
+    return 0 if metric_value_as_float.zero? || secondary_metric_value_as_float.zero?
+    search_rate = (metric_value_as_float / secondary_metric_value_as_float) * 100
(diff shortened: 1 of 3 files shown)


app/presenters/single_content_item_presenter.rb
```diff
@@ -146,11 +146,11 @@ def useful_yes_no_total
   end
 
   def on_page_search_rate
-    metric_value = self.total_searches
-    secondary_metric_value = self.total_pviews
+    metric_value_as_float = self.total_searches.delete(',').to_f
```
cbaines: How about using `@metrics['searches'][:value]` here, to avoid the formatting in the first place?

pmanrubia: @KyleMacPherson @cbaines Sorry I just merged and I didn't notice your comment. 

@KyleMacPherson could you please address @cbaines in a separate PR?

KyleMacPherson: @cbaines Nice!

KyleMacPherson:
[(1) Page data - fix % of users searched the page in the at a glance metrics](https://trello.com/c/TdGevWTq/764-1-page-data-fix-of-users-searched-the-page-in-the-at-a-glance-metrics)

PR #152, opened 2018-10-24
Refactor on_page_search_rate

KyleMacPherson:
As suggested in [this comment](https://github.com/alphagov/content-data-admin/pull/151/files#r227791900)

Use metric values to stop us from needing to format the human readable
version of them before using for calculations.

### DIFF
--- a/app/presenters/single_content_item_presenter.rb
+++ b/app/presenters/single_content_item_presenter.rb
@@ -145,11 +145,11 @@ private
   end
 
   def on_page_search_rate
-    metric_value_as_float = self.total_searches.delete(',').to_f
-    secondary_metric_value_as_float = self.total_pviews.delete(',').to_f
+    searches = @metrics['searches'][:value].to_f
+    pviews = @metrics['pviews'][:value].to_f
 
-    return 0 if metric_value_as_float.zero? || secondary_metric_value_as_float.zero?
-    search_rate = (metric_value_as_float / secondary_metric_value_as_float) * 100
+    return 0 if searches.zero? || pviews.zero?
+    search_rate = (searches / pviews) * 100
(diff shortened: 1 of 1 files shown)


app/presenters/single_content_item_presenter.rb
```diff
@@ -145,8 +145,8 @@ def useful_yes_no_total
   end
 
   def on_page_search_rate
-    metric_value_as_float = self.total_searches.delete(',').to_f
-    secondary_metric_value_as_float = self.total_pviews.delete(',').to_f
+    metric_value_as_float = @metrics['searches'][:value].to_f
+    secondary_metric_value_as_float = @metrics['pviews'][:value].to_f
```
cbaines: Can you get away with not calling `to_f` here? I'm wondering if it might be a number already.

cbaines: I'd also be tempted to change the variable names here as well, maybe you could use `searches`, or `searches_value` rather than `metric_value_as_float`, as I think that might make the following code easier to read?

KyleMacPherson: @cbaines it is an integer so we need to use to_f to avoid doing integer division. For example `9/5` returns `1` as opposed to `9.0/5.0` returning `1.8`.

Have changed the variable names, reads much better, thanks!

cbaines: Ah, that makes sense :+1:

KyleMacPherson:
[(1) Page data - fix % of users searched the page in the at a glance metrics](https://trello.com/c/TdGevWTq/764-1-page-data-fix-of-users-searched-the-page-in-the-at-a-glance-metrics)

PR #157, opened 2018-10-26
Use unique pageviews for search rate

KyleMacPherson:
#### What
[Trello card](https://trello.com/c/TdGevWTq/764-1-page-data-fix-of-users-searched-the-page-in-the-at-a-glance-metrics)
We currently use pageviews to calculate search rate, this PR changes it so that
we will use unique pageviews.

Some changes to the test data.

#### Why
Unique pageviews more accurately represents the idea of 'How many users searched per page'.

Changing the test data was necessary to make it closer to the kind of data that we will see in production.

---
#### Review Checklist
* [x] Changes in scope.
* [x] Added to trello card.

### DIFF
--- a/spec/features/single_content_item_spec.rb
+++ b/spec/features/single_content_item_spec.rb
@@ -42,7 +42,7 @@ RSpec.describe '/metrics/base/path', type: :feature do
 
     describe 'glance metrics section' do
       it 'renders glance metrics unique page views' do
-        expect(page).to have_selector '.glance-metric.upviews', text: '33'
+        expect(page).to have_selector '.glance-metric.upviews', text: '6,000'
       end
 
       it 'renders glance metric context for unique pageviews' do
@@ -50,7 +50,7 @@ RSpec.describe '/metrics/base/path', type: :feature do
       end
 
       it 'renders trend percentage for unique pageviews' do
-        expect(page).to have_selector '.upviews .app-c-glance-metric__trend', text: '+230.00%'
+        expect(page).to have_selector '.upviews .app-c-glance-metric__trend', text: '-40.00%'
       end
 
       it 'renders glance metric for satisfaction score' do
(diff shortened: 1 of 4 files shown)


spec/features/single_content_item_spec.rb
```diff
@@ -86,15 +86,15 @@
       end
 
       it 'renders context for page searches' do
-        expect(page).to have_selector '.searches .app-c-glance-metric__context', text: '0.4%'
+        expect(page).to have_selector '.searches .app-c-glance-metric__context', text: '4.0%'
```
theseanything: Should this be testing for two decimal places as per the design?

KyleMacPherson: Changed the data to something with 2 decimal places. Is this what you meant @theseanything

KyleMacPherson:
@theseanything `on_page_search_rate` is a private method so we can't unit test it, but we do have tests for the output in `single_content_item_spec.rb`

theseanything:
Ah right - in which case can we add unit tests for `searches_context`?

spec/presenters/single_content_item_presenter_spec.rb
```diff
@@ -46,6 +46,38 @@
     end
   end
 
+  describe '#searches_context' do
+    it 'shows the percentage of users who searched the page' do
+      current_period_data[:time_series_metrics] = [{ name: 'upviews', total: 100 }, { name: 'searches', total: 10 }]
+      subject = SingleContentItemPresenter.new(current_period_data, previous_period_data, date_range)
+      expect(subject.searches_context).to eq "10.0% of users searched from the page"
```
theseanything: Is the formatting always 2 decimal places?

spec/presenters/single_content_item_presenter_spec.rb
```diff
@@ -46,6 +46,38 @@
     end
   end
 
+  describe '#searches_context' do
+    it 'shows the percentage of users who searched the page' do
+      current_period_data[:time_series_metrics] = [{ name: 'upviews', total: 100 }, { name: 'searches', total: 10 }]
+      subject = SingleContentItemPresenter.new(current_period_data, previous_period_data, date_range)
```
theseanything: Is this not covered by line 16?

PR #159, opened 2018-10-26
Add pageviews_per_visit metric

KyleMacPherson:
#### What
[Trello card](https://trello.com/c/174c4iAe/758-2-page-data-add-pageviews-per-visit-metric)
Add pageviews per visit metric to single content item page

#### Why
So that users can measure how many times the page is viewed by an individual user in order to judge the performance of the page.

#### Screenshots
screen shot 2018-10-26 at 16 50 59

---
#### Review Checklist
* [ ] Changes in scope.
* [ ] Added to trello card.

### DIFF
--- a/app/views/metrics/show.html.erb
+++ b/app/views/metrics/show.html.erb
@@ -108,25 +108,25 @@
   <div class="govuk-grid-column-full section-performance">
     <h2 class="section-performance__header"><%= t ".section_headings.performance" %></h2>
 
-    <%= render 'metric_section', 
+    <%= render 'metric_section',
         metric_name: 'upviews',
-        total: @performance_data.total_upviews, 
+        total: @performance_data.total_upviews,
         short_context: nil,
         external_link: nil %>
 
     <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
 
-    <%= render 'metric_section', 
-        metric_name: 'pviews', 
-        total: @performance_data.total_pviews, 
+    <%= render 'metric_section',
+        metric_name: 'pviews',
+        total: @performance_data.total_pviews,
         short_context: nil,
         external_link: nil %>
 
     <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
 
-    <%= render 'metric_section', 
-        metric_name: 'searches', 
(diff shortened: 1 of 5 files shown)


app/views/metrics/show.html.erb
```diff
@@ -108,43 +108,54 @@
...
         external_link: nil %>
 
     <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
 
-    <%= render 'metric_section', 
-        metric_name: 'pviews', 
-        total: @performance_data.total_pviews, 
+    <%= render 'metric_section',
+        metric_name: 'pviews',
+        total: @performance_data.total_pviews,
```
cbaines: I'm going to try and split the other changes out from the functional changes to make this a bit easier to read.

cbaines: Done now in 35430a156ce5437d87dc20da72513ea222c4ffdf

theseanything:
Did we want to make `pageviews_per_visit` a metric available from the API?

KyleMacPherson:
@theseanything I spoke to @pmanrubia on friday afternoon about this story on Friday afternoon and we agreed that the most pragmatic approach for now was to make this a front end change for now.

spec/support/content_data_api.rb
```diff
@@ -211,7 +211,7 @@ def default_previous_single_page_payload(base_path, from, to)
             },
             {
               name: "pviews",
-              total: 30,
+              total: 30000,
```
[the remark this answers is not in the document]
KyleMacPherson: Thanks @cbaines. The motivation was to both create more realistic test data and to make sure we have checks around the way we format data. We had a bug, due to the way we use `number_with_delimiter`, that wasn't represented in our tests because the values in our test data was all <1000 which meant that no commas were included in the numbers. (The bug was fixed [here](https://github.com/alphagov/content-data-admin/pull/151)).

Sorry for the lack of explanation in this PR.

KyleMacPherson:
[(2) Page data - Add pageviews per visit metric](https://trello.com/c/174c4iAe/758-2-page-data-add-pageviews-per-visit-metric)

PR #160, opened 2018-10-29
Fix javascript errors

maxgds:
#### What
https://trello.com/c/2NUh4o8A/777-1-bug-fix-js
This shifts the call to the filter toggle js to make it exclusive to where it is needed, and removes 
the global call to application js as the gem covers this. It also adds some spacing to mobile
views.

#### Why
Fixes js errors and tidies up views on narrow screens.

---
#### Review Checklist
* [/ ] Added to trello card.


maxgds:
@alex-ju thanks for the tip! Could you take another look at it now, please?

PR #165, opened 2018-10-29
Fix table scroll

maxgds:
#### What
Apply the scroll to the table and not the table header
https://trello.com/c/0uFwHlct/782-1-about-this-data-detail-link-disappears-when-user-scrolls-horizontally

#### Why
So that users can access the show/hide mechanism after scrolling without having to scroll back

### DIFF
--- a/app/assets/stylesheets/components/_chart.scss
+++ b/app/assets/stylesheets/components/_chart.scss
@@ -5,8 +5,11 @@
   margin-bottom: 0;
 }
 
-.app-c-chart__table {
+.app-c-chart__table-wrapper {
   overflow: auto;
+}
+
+.app-c-chart__table {
 
   .govuk-details__summary-text {
     @include govuk-font(16);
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/components/_chart.scss
```diff
@@ -5,6 +5,10 @@
   margin-bottom: 0;
 }
 
+.app-c-chart__table-wrapper {
+  overflow: auto;
+}
+
 .app-c-chart__table {
   overflow: auto;
```
cbaines: Can this be removed now? It looks to me that it might be redundant with this change.

maxgds: yes, my mistake caused by breaking my branch and starting again. Now fixed.

maxgds:
@cbaines could you recheck, please?

app/assets/stylesheets/components/_chart.scss
cbaines: :+1:

cbaines:
> @cbaines could you recheck, please?

Yep, looks good to me :)

PR #168, opened 2018-10-30
Fix searches_per_visit_bug

KyleMacPherson:
### What?
We currently have a bug which raises an error when calculating searches per page
if we do not have any data for either pviews or upviews for a certain period. It was introduced in [this PR](https://github.com/alphagov/content-data-admin/pull/159)

This bug has raised it's head when we set the date range to a period longer than we
have data for. This is because we will not have any data for the previous period, which
we use to calculate trends and so we get the error which is raised by this bug:

'`NoMethodError: undefined method `zero?' for nil:NilClass'

### How?
This PR calls `.to_f` on the value field of pviews and upviews. This will convert nil to 0
which will stop the error being raised if we receive a nil value.

### Notes

If we do not have any value for a set period then we will set pageviews per visit to `0`. This is not strictly true but is currently the strategy we are using in other areas. This is something the product team are aware of and are considering.

---
#### Review Checklist
* [ ] Changes in scope.
* [ ] Added to trello card.

### DIFF
--- a/app/presenters/single_content_item_presenter.rb
+++ b/app/presenters/single_content_item_presenter.rb
@@ -196,7 +196,7 @@ private
   end
 
   def calculate_current_pageviews_per_visit
-    current = if @metrics['pviews'][:value].zero? || @metrics['upviews'][:value].zero?
+    current = if @metrics['pviews'][:value].to_f.zero? || @metrics['upviews'][:value].to_f.zero?
                 0
               else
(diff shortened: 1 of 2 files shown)


KyleMacPherson:
Agreed. Thanks @cbaines!

KyleMacPherson:
[(2) Page data - Add pageviews per visit metric](https://trello.com/c/174c4iAe/758-2-page-data-add-pageviews-per-visit-metric)

PR #169, opened 2018-10-30
Pagination

peteglondon:
#### What
Add pagination to the index page

#### Why
So the user can view more than the first page.

There is a related [PR in content-performance-manager](https://github.com/alphagov/content-performance-manager/pull/974)

Url to test: http://localhost:3230/content?date_range=last-month&document_type=&organisation_id=c36bd301-d0c5-4492-86ad-ee7843b8383b&page=4

##### Before
screen shot 2018-10-30 at 12 44 53

##### After

screen shot 2018-11-01 at 11 41 04

### DIFF
--- a/app/controllers/content_controller.rb
+++ b/app/controllers/content_controller.rb
@@ -14,7 +14,9 @@ private
     response = FindContent.call(content_params)
     ContentItemsPresenter.new(
       response[:results],
-      DateRange.new(content_params[:date_range])
+        DateRange.new(content_params[:date_range]),
+        response[:total_results],
+        params[:page]
     )
   end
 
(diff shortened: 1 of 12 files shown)


app/controllers/content_controller.rb
```diff
@@ -14,7 +14,9 @@ def get_content
     response = FindContent.call(content_params)
     ContentItemsPresenter.new(
       response[:results],
-      DateRange.new(content_params[:date_range])
+        DateRange.new(content_params[:date_range]),
```
pmanrubia: The indentation seems to be broken here.
We have increased the number of parameters that the Presenter is receiving. I am wondering about passing only two: `response` and `content_params`? Other can be inferred from these two parameters.

peteglondon:
[(5) Index page - Add pagination](https://trello.com/c/9SpR0P1I/662-5-index-page-add-pagination)

PR #179, opened 2018-11-05
Handle no comparison data

KyleMacPherson:
#### What
[Trello card](https://trello.com/c/a34qyfIg/786-2-page-data-display-no-data-when-there-isnt-comparison-data-available-for-a-previous-time-period)
Display 'no comparison data' when we have no previous data with which to compare current data.

Currently we display a trend percentage of 0% if we have no previous data.

#### Why
Displaying a 0% change for both an actual 0% change as well as an unknown change is incorrect and is likely to be confusing for end users who do not know which of the two scenarios that the 0% change would relate to.

### Before
screen shot 2018-11-05 at 12 13 43

### After
screen shot 2018-11-05 at 12 12 06

---
#### Review Checklist
* [x] Changes in scope.
* [x] Added to trello card.

### DIFF
--- a/app/presenters/single_content_item_presenter.rb
+++ b/app/presenters/single_content_item_presenter.rb
@@ -224,8 +224,8 @@ private
   end
 
   def calculate_trend_percentage(current_value, previous_value, metric_name)
-    return 'no comparison data' if incomplete_previous_data?(current_value, previous_value, metric_name)
-    
+    return if incomplete_previous_data?(current_value, previous_value, metric_name)
+
     previous_value[:value] <= 0 ? 0 : ((current_value[:value].to_f / previous_value[:value].to_f) - 1) * 100
   end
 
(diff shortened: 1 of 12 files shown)


KyleMacPherson:
[(2) Page data - Display 'no data' when there isn't comparison data available for a previous time period](https://trello.com/c/a34qyfIg/786-2-page-data-display-no-data-when-there-isnt-comparison-data-available-for-a-previous-time-period)

app/presenters/single_content_item_presenter.rb
```diff
@@ -213,16 +213,42 @@ def calculate_current_pageviews_per_visit
   end
 
   def calculate_previous_pageviews_per_visit
-    previous = if @previous_metrics['pviews'][:value].to_f.zero? || @previous_metrics['upviews'][:value].to_f.zero?
+    previous = if @previous_metrics['pviews'][:value].blank? || @previous_metrics['upviews'][:value].blank?
```
chao-xian: Great to see the changes come into shape. One last thing please: the `single_content_item_presenter.rb` we've now got more calculation methods hidden in a private method. The other private methods in here were relatively low complexity. Now that we have methods that have increased complexity, it looks like a good time to abstract them into a new class that does calculations. Basically the `calculate_` methods look like good candidates for this abstraction. We can also test those in isolation.

PR #192, opened 2018-11-08
Guard against missing data to avoid layout breaking

maxgds:
# What
https://trello.com/c/eJtCoZ3g/826-page-data-fix-formatting-of-metadata-when-we-have-no-value-for-primary-publishing-organisation
Guards against empty metadata items and renders nothing for blank fields

# Why
Layout currently breaks when the value is empty

# Screenshots

## Before
screen shot 2018-11-08 at 15 25 02

## After

screen shot 2018-11-08 at 15 25 18

### DIFF
--- a/app/views/components/_metadata.html.erb
+++ b/app/views/components/_metadata.html.erb
@@ -1,31 +1,37 @@
 <%
-    published_at ||= ""
-    last_updated ||= ""
-    publishing_organisation ||= ""
-    document_type ||= ""
-    base_path ||= ""
-    status ||= ""
+    published_at ||= nil
+    last_updated ||= nil
+    publishing_organisation ||= nil
+    document_type ||= nil
+    base_path ||= nil
+    status ||= nil
 %>
 
 <div class="app-c-metadata govuk-body govuk-body-xs">
-  <% if status != "" %>
+  <% if status %>
     <dl class="app-c-metadata__list">
       <dt class="app-c-metadata__title"><%= t ".labels.status" %></dt>
       <dd class="app-c-metadata__description govuk-!-font-weight-bold"><%= status %></dd>
     </dl>
   <% end %>
   <dl class="app-c-metadata__list">
-    <dt class="app-c-metadata__title"><%= t ".labels.published_at" %></dt>
-    <dd class="app-c-metadata__description"><%= published_at %></dd>
-    <dt class="app-c-metadata__title"><%= t ".labels.last_updated" %></dt>
-    <dd class="app-c-metadata__description"><%= last_updated %></dd>
+    <% if published_at %>
+      <dt class="app-c-metadata__title"><%= t ".labels.published_at" %></dt>
+      <dd class="app-c-metadata__description"><%= published_at %></dd>
+    <% end %>
+    <% if last_updated %>
+      <dt class="app-c-metadata__title"><%= t ".labels.last_updated" %></dt>
+      <dd class="app-c-metadata__description"><%= last_updated %></dd>
+    <% end %>
   </dl>
   <dl class="app-c-metadata__list">
-    <dt class="app-c-metadata__title"><%= t ".labels.publishing_organisation" %></dt>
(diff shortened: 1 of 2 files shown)


app/views/components/_metadata.html.erb
```diff
@@ -1,31 +1,42 @@
...
-    <dt class="app-c-metadata__title app-c-metadata__title--wrappable"><%= t ".labels.base_path" %></dt>
-    <dd class="app-c-metadata__description"><% if base_path != "" %>gov.uk<%= base_path %><% end %></dd>
+    <% if publishing_organisation %>
+      <dt class="app-c-metadata__title"><%= t ".labels.publishing_organisation" %></dt>
+      <dd class="app-c-metadata__description"><%= publishing_organisation %></dd>
+    <% end %>
+    <% if document_type %>
+      <dt class="app-c-metadata__title"><%= t ".labels.document_type" %></dt>
+      <dd class="app-c-metadata__description"><%= document_type %></dd>
+    <% end %>
```
cbaines: Everything must have a document type, and I guess we display the document type for everything, so this conditional is probably unnecessary.

app/views/components/_metadata.html.erb
```diff
@@ -1,31 +1,42 @@
...
+
+    <% if base_path %>
+      <dt class="app-c-metadata__title app-c-metadata__title--wrappable"><%= t ".labels.base_path" %></dt>
+      <dd class="app-c-metadata__description"><% if base_path != "" %>gov.uk<%= base_path %><% end %></dd>
+    <% end %>
```
cbaines: Given that this is for the page data page, which is accessed using a path, I don't think there is a case where the base path is missing.

app/views/components/_metadata.html.erb
```diff
@@ -1,31 +1,42 @@
...
   <% end %>
   <dl class="app-c-metadata__list">
-    <dt class="app-c-metadata__title"><%= t ".labels.published_at" %></dt>
-    <dd class="app-c-metadata__description"><%= published_at %></dd>
-    <dt class="app-c-metadata__title"><%= t ".labels.last_updated" %></dt>
-    <dd class="app-c-metadata__description"><%= last_updated %></dd>
+    <% if published_at %>
+      <dt class="app-c-metadata__title"><%= t ".labels.published_at" %></dt>
+      <dd class="app-c-metadata__description"><%= published_at %></dd>
+    <% end %>
```
cbaines: I think everything must have published_at time, but I'm not totally confident...

maxgds:
@cbaines I've reworked it to the items we're less sure will always be there - could you take another look?

app/views/components/_metadata.html.erb
```diff
@@ -1,31 +1,37 @@
...
-    <dt class="app-c-metadata__title app-c-metadata__title--wrappable"><%= t ".labels.base_path" %></dt>
-    <dd class="app-c-metadata__description"><% if base_path != "" %>gov.uk<%= base_path %><% end %></dd>
+    <% if publishing_organisation %>
+      <dt class="app-c-metadata__title"><%= t ".labels.publishing_organisation" %></dt>
+      <dd class="app-c-metadata__description"><%= publishing_organisation %></dd>
+    <% end %>
+      <dt class="app-c-metadata__title"><%= t ".labels.document_type" %></dt>
+      <dd class="app-c-metadata__description"><%= document_type %></dd>
+      <dt class="app-c-metadata__title app-c-metadata__title--wrappable"><%= t ".labels.base_path" %></dt>
+      <dd class="app-c-metadata__description"><% if base_path != "" %>gov.uk<%= base_path %><% end %></dd>
```
cbaines: The indentation is a bit off for the four lines above.

cbaines: Also, with respect to this line about the base_path, I'm not sure the if on the line will work the same, as it's testing for `if base_path != ""`, and the default is now `nil`? I'm not sure why that's necessary, but maybe the if can be removed?

maxgds:
Thanks, tweaked and merging.

PR #197, opened 2018-11-12
Fix last-month comparison data bug

KyleMacPherson:
### What?

[Trello card](https://trello.com/c/a34qyfIg/786-3-page-data-display-no-data-when-there-isnt-comparison-data-available-for-a-previous-time-period#)

Fix trends in the at a glance section of the single page data page for the `past-month` date range.

### Why?

At a glance trends currentl returns `no comparison data` when the
date period `last-month` is selected. This is because we checked for
complete data based on the number of days in each month being the
same. Obviously this is not the case.

This PR changes the logic for the `past-month` date range in order to
check that the previous month has data for each day in the month,
regardless of the length of the month and regardless of how many
days are in the current month.

# Screenshots
*If applicable add screenshots otherwise remove this section.*

## After

---
# Review Checklist
* [x] Changes in scope.
* [x] Added to Trello card.

### DIFF
--- a/app/presenters/single_content_item_presenter.rb
+++ b/app/presenters/single_content_item_presenter.rb
@@ -220,10 +220,7 @@ private
     return true if no_previous_timeseries?(previous_value)
     return incomplete_last_month_data?(previous_value) if @date_range.time_period == 'last-month'
 
-    current_timeseries_length = length_of_timeseries(current_value[:time_series])
-    previous_time_series_length = length_of_timeseries(previous_value[:time_series])
-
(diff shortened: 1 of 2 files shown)


KyleMacPherson:
[(3) Page data - Display 'no data' when there isn't comparison data available for a previous time period](https://trello.com/c/a34qyfIg/786-3-page-data-display-no-data-when-there-isnt-comparison-data-available-for-a-previous-time-period)

chao-xian:
🎉

PR #207, opened 2018-11-16
Show full selected date range in charts

peteglondon:
# What
Pad out the data so that the full selected date range is shown.

# Why
We want to show the selected date range in the
metrics charts.

# Background
The correct way to do this would be to get the x axis
of the chart to recognise that it is rendering dates,
then to give a a min and max values for the x axis.

I've tried the first part of this and it works fine.
However the ChartKick js has a bug in it.

When passing the min and max values to google charts,
it mangles the date and the google chart ends up with an
empty object.

We have a work around which is to pad out the data with dummy
values for all the dates in the date range.

This also affects the table.

# Screenshots (last 6 months)

## Before
before

## After
after

Trello: 
# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.

### DIFF
--- a/app/presenters/chart_presenter.rb
+++ b/app/presenters/chart_presenter.rb
@@ -35,20 +35,33 @@ class ChartPresenter
   end
 
   def keys
-    return [] unless time_series
-
-    time_series.map do |point|
-      point[:date].to_date.strftime("%m-%d")
-    end
+    padded_time_series.map { |h| h[:date] }
   end
 
   def values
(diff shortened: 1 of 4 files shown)


peteglondon:
[(3) Page data - Display full date range in charts & tables even when we have no data](https://trello.com/c/wmZNZGet/714-3-page-data-display-full-date-range-in-charts-tables-even-when-we-have-no-data)

PR #209, opened 2018-11-16
Use date range to query warehouse content endpoint

KyleMacPherson:
# What
Use date range to query warehouse content endpoint

# Why
We are using materialised views to render content from the content endpoint in our back end. The content end point is set up to use a `date_range` metric rather than `from` and `to` metrics. This makes sense as it will be much easier to identify the required view from a `date_range` param rather than having to do complex logic to determine the required view from `from` and `to` params.

# Notes
This PR will not make any difference to the functionality of our index page on it's own. We currently have the back end hard coded to only ever show data from the last 30 days. It does however need to be merged before the back end changes so as not to break existing functionality.

---
# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.

### DIFF
--- a/app/services/find_content.rb
+++ b/app/services/find_content.rb
@@ -10,9 +10,7 @@ class FindContent
   end
 
   def initialize(params)
-    range = DateRange.new(params[:date_range])
-    @from = range.from
-    @to = range.to
+    @date_range = params[:date_range]
     @organisation = params[:organisation_id]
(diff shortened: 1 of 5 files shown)


KyleMacPherson:
[(3) Index page - Make date range filters work on index page](https://trello.com/c/52Imsnnq/841-3-index-page-make-date-range-filters-work-on-index-page)

PR #216, opened 2018-11-19
Add help text to index page table

maxgds:
# What

https://trello.com/c/fOwe3KC3/840-2-index-page-add-help-text-to-headers-on-table

This is the initial implementation which puts an SVG on the page,
and gives it a title tag - this means text will appear on a prolonged hover.

This also adds styles to support the table when we put it into the
non-sortable mode and applies that mode.

# Why
Move closer to designs, allow users to see help text.

# Screenshots

## Before

screen shot 2018-11-19 at 14 59 46

## After

screen shot 2018-11-19 at 15 00 33

### DIFF
--- a/app/views/content/index.html.erb
+++ b/app/views/content/index.html.erb
@@ -9,6 +9,8 @@
         <span class="filters-control filters-control--show govuk-visually-hidden">
           <a href="" class="govuk-link filters-control__link filters-control__link--show">Show filters</a>&nbsp;&#x25BC;</span>
       </div>
+
+
       <div class="filter-form">
         <%= form_tag content_path, method: 'get', name: 'organisation-picker' do %>
           <%= render "components/time-select", {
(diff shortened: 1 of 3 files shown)


app/views/content/index.html.erb
```diff
@@ -84,18 +86,20 @@
...
-              <%= t.header("Content type", href: '#', sort_direction: nil) %>
-              <%= t.header("Unique pageviews", href: '#', sort_direction: 'descending') %>
-              <%= t.header("User satisfaction score", href: '#', sort_direction: nil) %>
-              <%= t.header("Searches from page", href: '#', sort_direction: nil) %>
+              <%# t.header("Page title", href: '#', sort_direction: nil) %>
+              <%= t.header("Page title") %>
+              <%= t.header("Content type") %>
+              <%= t.header("Unique pageviews<br>" "#{image_tag('question_mark_default.svg', alt: 'Help', title: "#{t "metrics.upviews.summary"}")}".html_safe) %>
+              <%= t.header("User satisfaction score<br>" "#{image_tag('question_mark_default.svg', alt: 'Help', title: "#{t "metrics.satisfaction.summary"}")}".html_safe) %>
+              <%= t.header("Searches from page<br>" "#{image_tag('question_mark_default.svg', alt: 'Help', title: "#{t "metrics.searches.summary"}")}".html_safe) %>
```
cbaines: Are the double quotes here `<br>" "#{` intentional? I'm guessing that it would work the same without them.

app/views/content/index.html.erb
```diff
@@ -84,18 +86,20 @@
     </div>
     <div class="govuk-grid-column-three-quarters">
       <div class="table-wrapper">
+        <%=  %>
```
cbaines: Maybe remove this?

app/views/content/index.html.erb
maxgds: They were, but you're right, they weren't needed.

app/views/content/index.html.erb
maxgds: fixed

PR #224, opened 2018-11-22
Improve y-axis on charts

maxgds:
- prevent chart displaying negative numbers when all data is 0s
- prevent chart displaying fractional numbers in labels
- set sensible ranges on charts depending on max value in data, including capping charts showing percentages at 100

# What
Improve y axis as per 
https://trello.com/c/gVCYMq4q/869-3-improve-vertical-axis-on-charts

# Why
More logical chart views

# Screenshots

## Before
screen shot 2018-11-22 at 14 39 19

## After
screen shot 2018-11-22 at 14 39 51

## Before
screen shot 2018-11-22 at 14 41 44

## After
screen shot 2018-11-22 at 14 41 52

## Before
screen shot 2018-11-22 at 14 44 40

## After
screen shot 2018-11-22 at 14 44 48

### DIFF
--- a/app/views/components/_chart.html.erb
+++ b/app/views/components/_chart.html.erb
@@ -5,8 +5,28 @@
   table_direction ||= "horizontal"
   from ||= false
   to ||= false
+  percent_metric ||= false
   rows ||= []
   keys ||= []
+
+  if percent_metric
+    maximum_y = 100
+  elsif rows.any?
+    row_maximums = rows.map { |row| row[:values].max }
+    overall_maximum = row_maximums.max
+
+    if overall_maximum == 0
+      # set max value to 3 if data is all 0 avoid negative numbers on Y axis
+      maximum_y = 3
+    elsif overall_maximum < 10
+      # for low numbers let line hit top to avoid a fractional max value
+      maximum_y = overall_maximum
+    else
+      # for higher numbers give breathing space to line
+      maximum_y = overall_maximum * 1.3
+    end
+  end
+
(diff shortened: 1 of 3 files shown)


app/views/components/_chart.html.erb
```diff
@@ -5,8 +5,30 @@
...
   keys ||= []
+
+  if percent_metric
+    maximum_y = 100
+  else
+    current_max = 0
+    rows.each do |r|
+      max = r[:values].max
+      current_max = max.to_i unless current_max >= max.to_i
+    end
```
cbaines: How about
```ruby
row_maximums = rows.map { |row| row[:values].max }
overall_maximum = row_maximums.max
```

maxgds: Updated, also added a default value so the comparisons don't try to compare against null if no data is passed. Would appreciate if you have time to give it another look, @cbaines

cbaines: That looks good. With the `-1` default value. That looks like it will lead the `maximum_y` to be set of `-1`, is that an ok value?

maxgds: The charts/tables deliberately don't render when there's no row data, so it's irrelevant what the value is in that case, but this stops the comparison from exploding against a null value.

cbaines: Ah, I see, how about checking for this case earlier then, perhaps:
```ruby
  elsif row.any?
    row_maximums = rows.map { |row| row[:values].max }
    overall_maximum = row_maximums.max
```

maxgds: updated

cbaines: Great :+1:

PR #238, opened 2018-11-27
Fix crash on single item page.

peteglondon:
# What
Replace nils with 0 when calculating max value for chart.

# Why
We now allow nulls in the data.

The chart component is trying to find the max value for each
row in the dataset. If any of these are nil, the .max call fails.

---
# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.

### DIFF
--- a/app/views/components/_chart.html.erb
+++ b/app/views/components/_chart.html.erb
@@ -12,7 +12,7 @@
   if percent_metric
     maximum_y = 100
   elsif rows.any?
-    row_maximums = rows.map { |row| row[:values].max }
+    row_maximums = rows.map { |row| row[:values].compact.max }
     overall_maximum = row_maximums.max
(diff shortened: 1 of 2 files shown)


app/views/components/_chart.html.erb
```diff
@@ -12,7 +12,7 @@
   if percent_metric
     maximum_y = 100
   elsif rows.any?
-    row_maximums = rows.map { |row| row[:values].max }
+    row_maximums = rows.map { |row| row[:values].map{ |i| i || 0 }.max }
```
cbaines: How about `.compact.max`?

maxgds:
Could you add a test to spec/components/chart_spec.rb to test for null data to make sure this doesn't break again in future?

PR #248, opened 2018-11-29
Add route to handle homepage on single item page

peteglondon:
# What
Add a route to handle the GOV.UK homepage

# Why
The current routing fails because the base_path to the GOV.UK homepage
is an empty string once you have removed the leading `/`

Trello https://trello.com/c/q9YG0Ftm/915-fix-bug-with-rendering-the-govuk-homepage-in-content-data-admin

there is a [related PR in content-performance-manager](https://github.com/alphagov/content-performance-manager/pull/1033)

# Screenshots
Only after is shown. 

## Before
!!!BOOM!!! 500 Error
## After
screen shot 2018-11-29 at 15 37 20

---
# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.

### DIFF
--- a/spec/routing/single_page_routing_spec.rb
+++ b/spec/routing/single_page_routing_spec.rb
@@ -1,0 +1,35 @@
+RSpec.describe '/metrics routing' do
+  it 'routes correctly with given base path' do
+    expect(get: '/metrics/long/base/path?date_range=past-30-days').to route_to(
+      controller: 'metrics',
+      action: 'show',
+      base_path: 'long/base/path',
+      date_range: 'past-30-days'
+    )
+  end
+
+  it 'generates a route for the given base_path' do
+    expect(get: metrics_path('/base/path', date_range: 'past-30-days')).to route_to(
+      controller: 'metrics',
+      action: 'show',
+      base_path: 'base/path',
+      date_range: 'past-30-days'
+    )
+  end
+
+  it 'routes correctly without a base path for the homepage' do
+    expect(get: '/metrics?date_range=last-month').to route_to(
+      controller: 'metrics',
+      action: 'show',
+      date_range: 'last-month'
+    )
+  end
+
+  it 'generates a route for the homepage' do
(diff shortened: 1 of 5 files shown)


peteglondon:
[Fix bug with rendering the GOV.UK homepage in content-data-admin](https://trello.com/c/q9YG0Ftm/915-fix-bug-with-rendering-the-govuk-homepage-in-content-data-admin)

spec/routing/single_page_routing_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  it 'generates a route for the given base_path' do
+    expect(get: metrics_path('/base/path', date_range: 'past-30-days')).to route_to(
+      controller: 'metrics',
+      action: 'show',
+      base_path: 'base/path',
+      date_range: 'past-30-days'
+    )
+  end
```
theseanything: Whats the difference between this example and the one above?

spec/routing/single_page_routing_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  it 'generates a route for the homepage' do
+    expect(get: metrics_path(date_range: 'past-30-days')).to route_to(
+      controller: 'metrics',
+      action: 'show',
+      date_range: 'past-30-days'
+    )
+  end
```
theseanything: Again duplicated example?

theseanything: We should have a test for `/metrics/` - without any query parameters.

spec/routing/single_page_routing_spec.rb
peteglondon: We are testing route generation here, the one above is testing route recognition.

spec/routing/single_page_routing_spec.rb
peteglondon: Same as above

PR #249, opened 2018-11-29
CSV export fix URL and change some headings

cbaines:
https://trello.com/c/yLoOoyUv/864-5-index-page-add-download-csv

### DIFF
--- a/app/presenters/content_items_csv_presenter.rb
+++ b/app/presenters/content_items_csv_presenter.rb
@@ -13,7 +13,9 @@ class ContentItemsCSVPresenter
   def csv_rows
     fields = {
       'Title' => raw_field(:title),
-      'URL' => raw_field(:row),
+      'URL' => lambda do |result_row|
+        url(result_row[:base_path])
+      end,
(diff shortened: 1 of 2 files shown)


app/presenters/content_items_csv_presenter.rb
```diff
@@ -13,15 +13,17 @@ def initialize(data_enum, search_params, document_types, organisations)
...
+        url(result_row[:base_path])
+      end,
       'Content Data Link' => lambda do |result_row|
         content_data_link(result_row[:base_path])
       end,
       'Document Type' => raw_field(:document_type),
-      'Upviews' => raw_field(:upviews),
-      'Satisfaction' => raw_field(:satisfaction),
+      'Unique Pageviews' => raw_field(:upviews),
+      'User Satisfaction score' => raw_field(:satisfaction),
```
theseanything: Should we keep casing consistent? /could we use the rails locales for metric names `config/locales/defaults/en.yml`? - so if a content designer changed the text we show that in the CSVs

PR #251, opened 2018-11-30
GTM javascript

erino:
# What
Track filter form submission data and window size in Google Tag Manager via the dataLayer.

# Why
We can't use custom javascript in Google Tag Manager

---
# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.


app/assets/javascripts/core/gtm.js
```diff
@@ -0,0 +1,56 @@
...
+    window.dataLayer.push({
+      event: 'filter-form',
+      field: 'organisation',
+      value: organisation.selectedOptions[0].text
+    });
+
+  };
+
+  function init() {
+    document.getElementById("filter-form").addEventListener("submit", filterFormSubmitListner, false);
```
alex-ju: Should we check for the element to exist before adding the event listener?

app/assets/javascripts/core/gtm.js
```diff
@@ -0,0 +1,56 @@
+(function() {
+
+  function filterFormSubmitListner(event) {
+    window.dataLayer = window.dataLayer || [];
+    var form = document.querySelector('form[name="organisation-picker"]');
+
+    var date_range = form.querySelector('input[name="date_range"]:checked').value;
```
alex-ju: Shall we check that these selectors return anything before pushing to dataLayer? as we did with the search term?

erino: Fixed in 42758de

app/assets/javascripts/core/gtm.js
erino: Fixed in 42758de

PR #254, opened 2018-11-30
Remove duplicate linking of application.js

erino:
`application.js` is linked twice, in `layout/application.html.erb` and also as part of the layout_for_admin GOV.UK publishing component.

https://github.com/alphagov/govuk_publishing_components/blob/81d7aaf/app/views/govuk_publishing_components/components/_layout_for_admin.html.erb#L21


erino:
Turns out this breaks `Chartkick`, so I need to do a bit more work

PR #257, opened 2018-12-03
GTM javascript v2

erino:
* Updates to GTM tracking to make custom events easier to use
* Make gtm.js script more resilient to elements not being present
* a few small style fixes.

### DIFF
--- a/app/assets/javascripts/core/gtm.js
+++ b/app/assets/javascripts/core/gtm.js
@@ -7,31 +7,25 @@
 
     var dateRangeTag = form.querySelector('input[name="date_range"]:checked');
     if(dateRangeTag && dateRangeTag.value.length > 0) {
-      window.dataLayer.push({
-        event: eventName,
-        field: 'dateRange',
-        value: dateRangeTag.value
-      });
+      var message = {};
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/core/gtm.js
```diff
@@ -3,41 +3,40 @@
...
+    var eventName = form.id;
 
-    var date_range = form.querySelector('input[name="date_range"]:checked').value;
-    if(date_range.length > 0) {
-      window.dataLayer.push({
-        event: 'filter-form',
-        field: 'date_range',
-        value: date_range
-      });
+    var dateRangeTag = form.querySelector('input[name="date_range"]:checked');
```
theseanything: Maybe `dateRangeTag` is the wrong variable name here - as` querySelector()` returns a DOM element.

theseanything: Same with the use of `Tag` suffix further down

erino: Fixed in 270a1e21dfa62a4c9d05f4735f6b3fd4a5e70f02

PR #258, opened 2018-12-03
Format document types

KyleMacPherson:
# What
[Trello card](https://trello.com/c/xMSR3zFQ/899-3-change-document-types-to-the-correct-case-in-the-filter-list-and-in-the-metadata-section)

Format document types so that some acronyms are properly capitalised.

# Why
In order to help end users more easily understand the document types.

# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.

### DIFF
--- a/app/presenters/filter_presenter.rb
+++ b/app/presenters/filter_presenter.rb
@@ -21,7 +21,7 @@ class FilterPresenter
              }]
     @document_types.each do |document_type|
       types.push(
-        text: document_type.try(:tr, '_', ' ').try(:capitalize),
+        text: document_type.humanize,
         value: document_type,
         selected: document_type == @search_parameters[:document_type]
       )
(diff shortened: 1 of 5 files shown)


app/helpers/document_type_formatter_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+    'Foi' => 'FOI',
+    'Hmrc' => 'HMRC',
+    'Html' => 'HTML',
+    'Maib' => 'MAIB',
+    'Raib' => 'RAIB',
+    'Utaac' => 'UTAAC'
+  }.freeze
+
+  def self.format(document_type)
+    document_type = document_type.try(:tr, '_', ' ').try(:capitalize)
```
erino: Would [`humanize`](
https://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-humanize) work here?

erino: and with `ActiveSupport::Inflector::Inflections#acronym` it will avoid the need for the next few lines

PR #259, opened 2018-12-03
Add user to db/seeds.rb

erino:
Now a user doesn't need to be manually created when running the
app in development mode.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,7 +1,21 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rails db:seed command (or created alongside the database with db:setup).
-#
-# Examples:
-#
-#   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
-#   Character.create(name: 'Luke', movie: movies.first)
+def seed
+  unless Rails.env.development? || ENV["RUN_SEEDS_IN_PRODUCTION"]
+    puts "Skipping because not in development"
+    return
+  end
+
+  if User.where(name: "Test user").present?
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -1,7 +1,21 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rails db:seed command (or created alongside the database with db:setup).
-#
-# Examples:
-#
-#   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
-#   Character.create(name: 'Luke', movie: movies.first)
+def seed
+  unless Rails.env.development? || ENV["RUN_SEEDS_IN_PRODUCTION"]
```
theseanything: This seems dangerous and currently unneeded?

erino: Yes, I copied it from https://github.com/alphagov/specialist-publisher/blob/master/db/seeds.rb#L2-L5 without really questioning

Fixed in c65785d5e8d64c5a2a5b4cc39ab9ce5e5bf81c1e

db/seeds.rb
```diff
@@ -1,7 +1,21 @@
...
+    return
+  end
+
+  gds_organisation_id = "af07d5a5-df63-4ddc-9383-6a666845ebe9"
+
+  user = User.create!(
+    name: "Test user",
+    organisation_content_id: gds_organisation_id,
+    organisation_slug: "government-digital-service",
+  )
```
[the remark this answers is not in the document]
erino: Yep, fixed in c65785d5e8d64c5a2a5b4cc39ab9ce5e5bf81c1e

PR #274, opened 2018-12-07
Change the default filter to `All Organisations`

peteglondon:
# What
Change the default filter to `All Organisations`

# Why
We have found from user research that users are having
problems finding content.

We think this is because of the organisation filter which defaults
to the users organisation.

# Screenshots

## Before
before

## After
after

---
# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.


peteglondon:
[(2) Make all orgs the default](https://trello.com/c/pPno3H5B/931-2-make-all-orgs-the-default)

PR #279, opened 2018-12-11
Add CSV download for a single metric

KyleMacPherson:
EDIT: DO NOT MERGE IN CURRENT STATE, WILL BREAK SOME GA METRICS

# What
Add CSV download for a single metric. This involves adding a new metric controller and renaming our existing metric controller to single page controller. This was necessary because using the splat operator to assign the base path param in the old metrics (now single page) controller did not allow us to extend the controller.

Following the pattern established by the csv download work for the index page I have created a new single page csv presenter and made use of some of the existing `export_to_csv` logic.

# Why
In order to let publishers or editors be able to inspect individual metric data for a page in a spreadsheet.

# Screenshots
screen shot 2018-12-11 at 16 26 13

---
# Review Checklist
* [ ] Changes in scope.
* [x] Added to Trello card.


app/presenters/metrics_csv_presenter.rb
```diff
@@ -0,0 +1,28 @@
+require 'csv'
+
+class MetricsCSVPresenter
```
pmanrubia: Can you add test for this class?

spec/support/content_data_api.rb
```diff
@@ -119,6 +119,12 @@ def content_data_api_has_document_types
         stub_request(:get, url).to_return(status: 200, body: body)
       end
 
+      def content_data_api_has_csv
```
pmanrubia: We should be explicit when setting the fixture for a test. Could you please extract the body?

spec/features/csv_download_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    GDS::SSO.test_user = build(:user)
+    stub_metrics_page(base_path: 'base/path', time_period: :past_30_days)
+    visit '/single_page/base/path'
+    content_data_api_has_csv
+  end
+
+  it 'downloads csv' do
+    visit '/single_page/base/path?date_range=past-30-days'
+    click_link 'Download Unique pageviews CSV'
+    expect(CSV.parse(page.body)).to eq [%w(Date Value), [date2, "1000"], [date3, "2000"], [date1, "3000"]]
```
pmanrubia: It is not clear where do 1000, 2000 and 3000 comes from. What about being explicit when setting up the fixture?

KyleMacPherson: 👍

spec/support/content_data_api.rb
KyleMacPherson: 👍

app/presenters/metrics_csv_presenter.rb
KyleMacPherson: 👍

theseanything:
[(5) Add more fields in index page CSV download](https://trello.com/c/lyOMgJds/898-5-add-more-fields-in-index-page-csv-download)

spec/support/content_data_api.rb
```diff
@@ -119,12 +119,6 @@ def content_data_api_has_document_types
         stub_request(:get, url).to_return(status: 200, body: body)
       end
 
-      def content_data_api_has_csv
```
pmanrubia: We should be explicit when setting up the tests. It is better not to hide the response with the method, so it can be reused in different places:

I would suggest something like:

```ruby
  def content_data_api_has_csv(columns: [], values: [])
  end
```

And then use it with:

```ruby
content_data_api_has_csv(
  columns: %w(Date Value), 
  values: [['10-10-2018',1000],[['11-10-2018',1100]])
```
To be consistent with what we already have, I would put this method in `content_data_api`

KyleMacPherson: Good idea! 👍

app/views/content/index.html.erb
```diff
@@ -20,28 +20,33 @@
...
               {
                 value: "past-year",
-                text: t("metrics.show.time_periods.past-year.leading"),
+                text: t("single_page.show.time_periods.past-year.leading"),
                 hint_text: "#{(1.year.ago).strftime("%-d %B %Y")} to #{Date.yesterday.strftime("%-d %B %Y")}"
+              },
+              {
+                value: "past-2-years",
+                text: t("single_page.show.time_periods.past-2-years.leading"),
+                hint_text: "#{(2.years.ago).strftime("%-d %B %Y")} to #{Date.yesterday.strftime("%-d %B %Y")}"
```
cbaines: Is this change intentional? I remember something about this time range being removed, but I haven't checked.

KyleMacPherson: Ah well spotted, thanks Chris.

app/controllers/metrics_controller.rb
```diff
@@ -1,22 +1,18 @@
+require 'csv'
```
cbaines: This looks redundant, as `CSV` isn't used in this file.

app/presenters/single_content_item_presenter.rb
```diff
@@ -127,6 +127,12 @@ def link_text(metric_name)
     I18n.t("metrics.#{metric_name}.title").try(:downcase)
   end
 
+  def csv_url(metric_name)
+    from = @date_range.from
+    to = @date_range.to
+    "/metrics#{base_path}?metric_name=#{metric_name}&from=#{from}&to=#{to}.csv"
```
cbaines: Perhaps just remove the `.csv` bit from the end here? I don't think it adds much.

lib/gds_api/content_data_api.rb
```diff
@@ -64,7 +64,7 @@ def content_items_url(date_range, organisation_id, document_type, page, page_siz
   end
 
   def single_page_url(base_path, from, to)
-    "#{content_data_api_endpoint}/single_page/#{base_path}#{query_string(from: from, to: to)}"
+    "#{content_data_api_endpoint}/metrics/#{base_path}#{query_string(from: from, to: to)}"
```
cbaines: Is this change right? It seems odd to me.

pmanrubia:
I am closing this PR. It is difficult to follow as the commits are not clear and don't seem to be green. Also Kyle is not in the team any more to ask for clarifications.

PR #292, opened 2018-12-24
Fix link to Manuals Publisher for manual_section

chao-xian:
# What/Why

Links to edit `manual_sections` in Manuals Publisher require the
`content_id` of it's parent `manual`. This is now sent as
`parent_content_id` from the API.

# Screenshots

## Before
screen shot 2018-12-24 at 11 06 22 am

## After
screen shot 2018-12-24 at 11 06 36 am

---
# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.

### DIFF
--- a/app/helpers/external_links_helper.rb
+++ b/app/helpers/external_links_helper.rb
@@ -1,12 +1,16 @@
 module ExternalLinksHelper
-  def edit_url_for(content_id:, publishing_app:, base_path:, document_type:)
+  def edit_url_for(content_id:, publishing_app:, base_path:, document_type:, parent_content_id: '')
     case publishing_app
     when 'whitehall'
       "#{external_url_for('whitehall-admin')}/government/admin/by-content-id/#{content_id}"
     when 'publisher'
       "#{external_url_for('support')}/content_change_request/new"
     when 'manuals-publisher'
-      "#{external_url_for('manuals-publisher')}/manuals/#{content_id}"
+      if document_type == 'manual'
+        "#{external_url_for('manuals-publisher')}/manuals/#{content_id}"
+      elsif document_type == 'manual_section'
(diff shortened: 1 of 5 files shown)


chao-xian:
[(2) 'Edit in manuals publisher' link not working](https://trello.com/c/nBPI9QzB/935-2-edit-in-manuals-publisher-link-not-working)

app/helpers/external_links_helper.rb
```diff
@@ -1,12 +1,16 @@
 module ExternalLinksHelper
-  def edit_url_for(content_id:, publishing_app:, base_path:, document_type:)
+  def edit_url_for(content_id:, publishing_app:, base_path:, document_type:, parent_content_id: '')
```
KyleMacPherson: I would expect the default value to be `nil` here in most cases. Is there a benefit to having this default value as an empty string?

spec/presenters/single_content_item_presenter_spec.rb
```diff
@@ -230,7 +230,8 @@
         content_id: 'content-id',
         publishing_app: 'whitehall',
         base_path: '/the/base/path',
-        document_type: 'news_story'
+        document_type: 'news_story',
+        parent_content_id: nil
```
KyleMacPherson: I think the value here should be consistent with the default value. If the default value is `''` then I don't think we can ever have `nil` here right?

app/helpers/external_links_helper.rb
chao-xian: It's easier to receive an empty string than a nil via the API to then deal with  any `if nil` situations. Setting it to default to the same feels more consistent.

spec/presenters/single_content_item_presenter_spec.rb
chao-xian: Eek good spot.

chao-xian: Fixed

PR #318, opened 2019-01-15
Restrict /dev route to development and integration environments

chao-xian:
# What
Our frontend dev uses `/dev` for testing and development of components and styles as part of their development workflow but this route is live on production. This change sets it to 404 on any other environment but development and integration.

# Why
Users shouldn't be seeing this page.

---
# Review Checklist
* [ ] Changes in scope.


chao-xian:
[(1) Restrict the /dev page](https://trello.com/c/39Nr5ije/1029-1-restrict-the-dev-page)

PR #319, opened 2019-01-16
Reduce font size of headers on index page

maxgds:
# What
Resize headings
https://trello.com/c/qHoeN2OC/1028-1-make-all-headings-in-the-filter-column-and-the-sentence-summary-font-size-16px
PR also adjusts spacing so the text in the left and right columns align properly.

# Why
Match designs, fit more on screen

# Screenshots

## Before
screen shot 2019-01-16 at 10 53 12

## After
screen shot 2019-01-16 at 10 53 26

### DIFF
--- a/app/assets/stylesheets/content/_index.scss
+++ b/app/assets/stylesheets/content/_index.scss
@@ -7,7 +7,7 @@
   }
 
   .table-header {
-    @include govuk-font(19);
+    @include govuk-font(16);
     @include media(mobile) {
       @include govuk-responsive-padding(3, "left");
     }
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/content/_index.scss
```diff
@@ -18,11 +18,13 @@
...
 
   .table-header__param {
-    @include govuk-font($size: 19, $weight: bold);
+    @include govuk-font($size: 16, $weight: bold);
   }
 
   .govuk-table {
     background-color: $white;
+    // re-align text content of left and right columns
+    //margin-top: -1px;
```
cbaines: Maybe remove this?

maxgds: @cbaines Commenting it out was accidental - it's needed to make the two columns look right side by side (I was checking the effect in different browsers and accidentally left it commented out when I pushed.) Happy for it to go in still?

cbaines: Yeah, sure :+1:

PR #328, opened 2019-01-21
Simplify single item presenter part 1

chao-xian:
# What/Why
The SingleItemPresenter is doing much, much more than returning presentation data to the view. This is a first couple of steps to simplify some of the work.

---
# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.


chao-xian:
[(5) Simplify Single Content Item Presenter](https://trello.com/c/w3Wj2YGf/1046-5-simplify-single-content-item-presenter)

app/models/metric.rb
```diff
@@ -0,0 +1,22 @@
+class Metric
```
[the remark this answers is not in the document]
chao-xian: 🤦‍♂️... of course

spec/presenters/single_content_item/metadata_spec.rb
```diff
@@ -0,0 +1,69 @@
+RSpec.describe 'Metadata from single_content_item_presenter' do
```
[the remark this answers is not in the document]
chao-xian: Thanks that's a better suggestion. Fixed in:

https://github.com/alphagov/content-data-admin/pull/328/commits/d7c83f5158134a713ed502a091f0ae666f474d89

and

https://github.com/alphagov/content-data-admin/pull/328/commits/69391eaa0eed9283e48792ce7885f13cbfb0ceff

spec/models/metric_spec.rb
```diff
@@ -0,0 +1,50 @@
+RSpec.describe Metric do
+  let(:metric_data) do
```
pmanrubia: Would you mind moving the `let` closer to the scope describe where it is used? Would it be also possible to move it below the `it` so we prevent scrolling when reading the test?

spec/presenters/single_content_item/single_content_item_presenter_spec.rb
```diff
@@ -0,0 +1,238 @@
+RSpec.describe SingleContentItemPresenter do
```
pmanrubia: Do we need the folder `/single_content_item_presenter`?

spec/models/metric_spec.rb
chao-xian: Done https://github.com/alphagov/content-data-admin/pull/328/commits/c3e2177aa0f01deae437b8b82a96ac961cfd4fc5

spec/presenters/single_content_item/single_content_item_presenter_spec.rb
chao-xian: Sorry forgot to remove this mid-edits. Done https://github.com/alphagov/content-data-admin/pull/328/commits/41d4a7f60abb0a8c921c44ec923654fc1e51a164

spec/presenters/single_content_item/single_content_item_presenter_spec.rb
```diff
@@ -98,6 +98,104 @@
...
+
+      expect(subject.total_pdf_count).to eq('5,000')
+    end
+
+    it 'correctly formats number for a small value' do
+      current_period_data[:edition_metrics] = [{ name: 'pdf_count', value: 50 }]
+
+      expect(subject.total_pdf_count).to eq('50')
+    end
+  end
```
cbaines: There's quite a lot of code here. Given that it's common functionality for different fields, are there any Rspec features that would allow to express the same tests more concisely?

spec/presenters/single_content_item/single_content_item_presenter_spec.rb
```diff
@@ -98,6 +98,104 @@
...
+
+  describe '#total_feedex' do
+    it 'correctly formats number for a value in the millions' do
+      current_period_data[:time_series_metrics] = [{ name: 'feedex', total: 1_000 }, { name: 'upviews', total: '0' }, { name: 'pviews', total: 0 }]
+
+      expect(subject.total_feedex).to eq('1,000')
+    end
+
+    it 'correctly formats number for a small value' do
+      current_period_data[:time_series_metrics] = [{ name: 'feedex', total: 1 }, { name: 'pviews', total: 5 }, { name: 'upviews', total: 0 }]
```
cbaines: In this test, and some of the tests above, I think some of the test data in `current_period_data[:time_series_metrics]` is redundant, e.g. the `pviews` value here. If so, it would be good to remove it to make the tests clearer.

chao-xian: There's some code in the presenter tied up to the stuff in the private section whenever the presenter is initialised (`assign_pageviews_per_visit`). It always expects these values, but I've not had a chance to refactor that area yet and looked worth tackling in another PR.

app/presenters/single_content_item_presenter.rb
```diff
@@ -169,24 +164,6 @@ def metadata
...
-    end
-
-    edition_metrics.each do |metric|
-      metrics[metric[:name]] = {
-        'value': format_metric_value(metric[:name], metric[:value]),
-        'time_series': nil
-      }
-    end
-    metrics
-  end
```
cbaines: Rather than extracting this out in to a class method in a new class, I think it might be possible to push this out of the SingleItemPresenter entirely, and that may end up being simpler.

I think this is the wider context for this class: https://github.com/alphagov/content-data-admin/blob/4a88549f353941a57db1ab1c6f386600d7d9d76b/app/controllers/metrics_controller.rb#L9-L16

It might be neat to try and thread this transformation currently in `parse_metrics` above in to the `FetchSinglePage` class. The already transformed data could then be passed in to the `SingleContentItem` presenter, in the form that it needs. I haven't tried this, but it seems like it could work.

What do you think?

spec/presenters/single_content_item/single_content_item_presenter_spec.rb
theseanything: Re: my comment about creating generalised `formatted_metric` function
That would minimise tests here and write simpler tests for `MetricsFormatterHelper:: format_metric_value` which would ensure the correct format.

app/presenters/single_content_item_presenter.rb
pmanrubia: @cbaines I really like this idea. Although, I would create a separate PR for it.

spec/support/shared/trend_percentages.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+
+  def incomplete_previous_data
+    previous_month_time_series = []
+    (2.months.ago.to_date.beginning_of_month + 10.days..2.months.ago.to_date.end_of_month).each do |date|
+      previous_month_time_series << { date: date.strftime, value: 100 }
+    end
+    previous_month_time_series
+  end
+end
```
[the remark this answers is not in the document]
pmanrubia: Yes, SingleContentItemPresenter is doing too much. The scope for this PR is just an improvement over what we already have, not the final view. I am keen to keep the scope of the PR small, and continue adding PRs to have a proper presenter for the single page.

[I don't think I was clear enough on my feedback]([url](https://github.com/alphagov/content-data-admin/pull/328#discussion_r249720892)). Extracting Shared Examples was a preparatory commit towards extracting the behaviour out.

chao-xian:
[Refactorings Content Data Admin (1)](https://trello.com/c/iQaf7Nwv/1084-refactorings-content-data-admin-1)

app/presenters/single_content_item_presenter.rb
```diff
@@ -9,47 +9,42 @@ def initialize(current_period_data, previous_period_data, date_range)
...
 
   def total_pdf_count
-    number_with_delimiter @metrics['pdf_count'][:value]
+    number_with_delimiter value_for('pdf_count')
   end
 
   def pageviews_per_visit
-    number_with_delimiter @metrics['pageviews_per_visit'][:value]
+    number_with_delimiter value_for('pageviews_per_visit')
   end
```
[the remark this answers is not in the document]
chao-xian: Thanks for the good suggestion - [done](https://github.com/alphagov/content-data-admin/pull/328/commits/169989ee9fcee33ace6daa3fe98a069ad6d0edc6)

spec/presenters/single_content_item/single_content_item_presenter_spec.rb
chao-xian: I've taken Sean's [suggestion and removed the repetitive tests](https://github.com/alphagov/content-data-admin/pull/328/commits/169989ee9fcee33ace6daa3fe98a069ad6d0edc6).

app/presenters/single_content_item_presenter.rb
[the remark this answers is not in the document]
theseanything: Ah sorry, I think I've steered you down a wrong road. 

My original thinking was reducing each of the `total_<metric_name>` to a *single* function (to return the formatted value for any given metric name). 

However looking at it again `MetricsFormatterHelper::format_metric_value` is used in a bunch of other places (chart values, CSV values), so unsure of the impact of adding `number_with_delimiter`.

I think it probably best to leave these changes out - and take another look in a separate story.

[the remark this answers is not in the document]
chao-xian: Yeah I encountered its use in other places and took that into account.

PR #330, opened 2019-01-22
Implement accessibile autocomplete component for dropdowns

maxgds:
# What
https://trello.com/c/vbIFFSjh/1038-3-implement-basic-accessible-autocomplete-and-check-for-ie11-issues
This PR implements the accessible autocomplete component.
As agreed with Jeremy and Mark, when a user first sees the page (or when they clear their filters) we insert a blank option at the start so that the dropdown works as a dropdown, revealing the full content list. On submission we detect a hidden parameter the form sends and then auto-select their choice, or default to the "All document types" or "All organisations" option. This is because a selected item looks like a dropdown with just one option in it and a user who hasn't seen the full list may not consider deleting the content. 

This means that we are having to catch "all" as a document type where we didn't previously.

# Why
The autocomplete will allow users to find their organisation or document type more easily by typing if they know its name or abbreviation.

# Screenshots

## Before
screen shot 2019-01-23 at 15 50 28

## After
screen shot 2019-01-23 at 15 49 50

### DIFF
--- a/app/assets/stylesheets/content/_index.scss
+++ b/app/assets/stylesheets/content/_index.scss
@@ -74,6 +74,10 @@
   .filter-form {
     @include govuk-responsive-padding(3, "left");
 
+    .govuk-label--s {
+      @include govuk-font($size: 16, $weight: bold);
+    }
+
     .govuk-radios__hint {
       @include govuk-font(16);
     }
--- a/app/presenters/filter_presenter.rb
+++ b/app/presenters/filter_presenter.rb
@@ -18,27 +18,58 @@ class FilterPresenter
   end
 
   def document_type
-    find_document_type[:text]
+    find_selected_document_type || ['', '']
+  end
+
+  def document_type_id
+    find_selected_document_type[1]
+  end
+
+  def document_type_name
+    document_type[0]
+  end
+
+  def selected_document_type(params)
+    if params[:submitted].blank?
+      ['', '']
+    else
+      document_type
+    end
+  end
+
+  def organisation
+    find_selected_org || ['', 'all']
   end
 
   def organisation_name
-    find_selected_org[:text]
+    org_name = find_selected_org[0]
+    if org_name == ""
+      "All organisations"
+    else
+      org_name
+    end
+  end
+
+  def organisation_id
+    find_selected_org
+  end
+
+  def selected_organisation(params)
+    if params[:submitted].blank?
+      ['', '']
+    else
+      organisation
+    end
   end
 
   def document_type_options
-    types = [{
-               text: 'All document types',
-               value: '',
-               selected: @search_parameters[:document_type] == ''
-             }]
-    @document_types.each do |document_type|
-      types.push(
-        text: document_type[:name].humanize,
-        value: document_type[:id],
-        selected: document_type[:id] == @search_parameters[:document_type]
-      )
+    @document_type_options ||= begin
+      types = [['', 'all'], ['All document types', 'all']]
+      @document_types.each do |document_type|
+        types.push([document_type[:name], document_type[:id]])
+      end
+      types
     end
-    types
   end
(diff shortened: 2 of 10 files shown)


app/presenters/filter_presenter.rb
```diff
@@ -47,29 +62,25 @@ def organisation_options
...
+          [name, org[:id]]
         end
     end
   end
 
 private
 
   def find_selected_org
-    organisation_options.find { |o| o[:selected] }
+    organisation_options.select { |o| o[1] == @search_parameters[:organisation_id] }[0] || ['All organisations', 'all']
```
chao-xian: Are we skipping the first option because it's blank? Can we skip it before we do the select? Also with `[0]` would `[:name]` not work? Apologies for not seeing it clearly. Same for the document type select.

app/views/content/index.html.erb
```diff
@@ -1,5 +1,18 @@
 <% content_for :title, @presenter.title %>
 <% content_for :page_class, "content-data__index" %>
+<%
+  if params[:submitted]
```
chao-xian: Can we move these out of the view?

app/presenters/filter_presenter.rb
maxgds: We've built an array of arrays:
[[name1, id1],[name2, id2]] 
because this is the format the component needs.
We're querying that to find the option with the right value, then returning that name and id as an array. The [0] is to get the contents of the container array, the o[1] is to query against the 2nd part of the array where the id is as it's not a hash at this point. Happy to change it if there's a clearer way to write it.

app/views/content/index.html.erb
maxgds: I tried but couldn't access the "submitted" param from the presenter - happy to fix if shown how.

app/presenters/filter_presenter.rb
maxgds: I've confused myself, now. Revisiting this.

app/views/content/index.html.erb
chao-xian: You'd have to pass the params to the presenter, but I'm not sure if that's a good idea myself.

app/presenters/filter_presenter.rb
maxgds: Okay, so I was using select, which returned an array, so I've changed that and been able to lose the [0]. This needs to return a slice of the array (which is an array itself) so I couldn't use [:name] here.

app/views/content/index.html.erb
maxgds: I've updated this - I no longer make the blank selectable at the top conditional, and I pass the presenter the page params and work with it to figure out whether to select the blank option or the explicit "All" option.

app/assets/stylesheets/content/_index.scss
```diff
@@ -74,6 +74,10 @@
   .filter-form {
     @include govuk-responsive-padding(3, "left");
 
+    .govuk-label--s {
```
alex-ju: If this needs to be smaller than the `--s` size of a label (which is currently `19px`) shall we define this as a `.app-label--xs` to know that it's smaller than the `--s` and it's being defined at the app level?

I do get this is an overwrite, being applied to the hints too, where you most probably won't be able to pass a class right now, so I get if this is not possible at the moment.

app/views/content/index.html.erb
```diff
@@ -11,8 +11,10 @@
       </div>
 
 
+
       <div class="filter-form">
         <%= form_tag content_path, method: 'get', name: 'organisation-picker', id: 'filter-form' do %>
+          <input type="hidden" name="submitted" value="true" >
```
alex-ju: This seems a bit flaky, but I do get the need to separate `All` from the other options.

PR #352, opened 2019-01-30
Add reading time metric to interface

maxgds:
# What
https://trello.com/c/fNfu8Vl7/1122-3-add-reading-time-to-interface-front-end

This shouldn't be merged because we don't have the reading time metric yet, so it is faked here, but this PR:

- Adds reading time to the view
- Tests for it
- Adjusts CSS to make layout accommodate the new column

# Why
Helps users understand word count in a less abstract way

# Screenshots

## Before
screen shot 2019-01-30 at 13 15 27

## After
screen shot 2019-01-30 at 13 14 53

### DIFF
--- a/config/locales/defaults/en.yml
+++ b/config/locales/defaults/en.yml
@@ -102,6 +102,20 @@ en:
         Lots of words on a page can make content difficult for users to read online. The longer the
         page is, the less usable it is likely to be. You could investigate whether this content could
         be shorter. Otherwise, does it have headings and bulletpoints to help users scan?
+    reading_time:
+      title: 'Reading time'
+      short_title: 'Reading time'
+      summary: 'Typical time to read all %{number_of_words} words'
+      context: ''
+      unit: ''
+      data_source: none
+      external_link: ''
+      about_title: 'About reading time'
+      about: >
+        Reading time is calculated by dividing the word count for the page by an average reading speed
+        of 200 words per minute. The time it takes to read the page is rounded up to the nearest minute.
+        This calculation doesn’t take into account how complicated the content is or any of the page’s
+        attachments.
     pdf_count:
       title: 'Number of PDFs'
       short_title: 'Number of PDFs'
@@ -112,7 +126,7 @@ en:
       external_link: ''
       about_title: 'About number of PDFs'
       about: >
-        Compared with HTML content, information published in a PDF is harder to find, use and 
+        Compared with HTML content, information published in a PDF is harder to find, use and
         maintain. PDFs can often be bad for accessibility and rarely comply with open standards.
         Several PDFs on one page may indicate that too many users needs are being met by them.
         <a href="https://gds.blog.gov.uk/2018/07/16/why-gov-uk-content-should-be-published-in-html-and-not-pdf/">Read more about why to avoid PDFs</a>.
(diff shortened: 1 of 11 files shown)


chao-xian:
[(2) Calculate reading time metric (backend)](https://trello.com/c/24kx1Ytt/1107-2-calculate-reading-time-metric-backend)

theseanything:
Waiting for comments in #content-performance-manager/pull/1129 to be addressed

app/helpers/duration_formatter_helper.rb
```diff
@@ -0,0 +1,6 @@
+module DurationFormatterHelper
```
theseanything: Could this go in the MetricsFormatterHelper module?

app/helpers/duration_formatter_helper.rb
```diff
@@ -0,0 +1,6 @@
...
+  def format_duration(minutes)
+    dur = ActiveSupport::Duration.build(minutes * 60)
+    Time.at(dur).utc.strftime("%kh %-Mm").strip
```
theseanything: For consistency, we should use the `Time.to_s` method with and use a formatter specified in the `config/time_formats.rb`. See here for [docs](https://api.rubyonrails.org/classes/Time.html#method-i-to_formatted_s)

theseanything: What's the call to `strip` for?

config/locales/defaults/en.yml
```diff
@@ -102,6 +102,20 @@ en:
...
+      context: ''
+      unit: ''
+      data_source: none
+      external_link: ''
+      about_title: 'About reading time'
+      about: >
+        Reading time is calculated by dividing the word count for the page by an average reading speed
+        of 200 words per minute. The time it takes to read the page is rounded up to the nearest minute.
+        This calculation doesn’t take into account how complicated the content is or any of the page’s
+        attachments.
```
theseanything: @pollygreen have you had a chance to look at this copy?

theseanything:
@chao-xian there have also been some changes to the organisation of the request stubs - let me know if you want me to talk you through it!

config/locales/defaults/en.yml
maxgds: @pollygreen wrote it!

app/helpers/duration_formatter_helper.rb
chao-xian: The `%k` and `%-M` formats remove leading zeros but pad with spacing (which don't fit the design and is also harder to test). `.strip` removes leading (and trailing) whitespace.

app/helpers/duration_formatter_helper.rb
chao-xian: Done c13e348

app/helpers/duration_formatter_helper.rb
chao-xian: Done df89760

theseanything: Would using ""%-kh %-Mm" work? The `-` should prevent padding with space.

chao-xian: Done 05087bc

PR #393, opened 2019-02-25
Unescape foreign characters on slugs

maxgds:
# What
Fix display bug by decoding unusual characters encoded in slugs
https://trello.com/c/EHrM18LF/1548-bug-some-foreign-language-slugs-stretch-the-page-title-column-width-in-the-table-and-link-to-page-not-found

# Why
Page currently stretches and slugs are incomprehensible

# Screenshots

## Before
screen shot 2019-02-25 at 09 16 23

## After
screen shot 2019-02-25 at 09 16 38

### DIFF
--- a/app/views/content/_page_title.html.erb
+++ b/app/views/content/_page_title.html.erb
@@ -5,4 +5,4 @@
     'gtm-item-document-type': item.raw_document_type
   } %>
 <br>
-<span class="base-path">/<%= item.base_path %></span>
+<span class="base-path">/<%= CGI::unescape(item.base_path) %></span>


app/views/content/_page_title.html.erb
```diff
@@ -5,4 +5,4 @@
     'gtm-item-document-type': item.raw_document_type
   } %>
 <br>
-<span class="base-path">/<%= item.base_path %></span>
+<span class="base-path">/<%= CGI::unescape(item.base_path) %></span>
```
pmanrubia: Not sure about scaping them in our application. The `base_paths` seems to be broken for the Content Items in `PublishingAPI` as well. 

I would rather fixing them in `PubishingAPI`?

Any thoughts @chao-xian @cbaines?

cbaines: I'm unsure what goes on in the Publishing API, but we can check. Even if the Publishing API sends encoded URLs, it's probably better to decode them all as part of the ETL process, rather than in the view.

maxgds:
Closing since it's not the ideal approach.

pmanrubia:
@maxgds @cbaines 

I have tried to escape the characters before storing them, but there are some scenarios that are not playing very well (for example, when opening the base_paths and parsing it) I would rather exploring it in detail in other stories, and do a short term fix for this issue.

I am now keen to merge this PR, hopefully you both are also happy with it.

maxgds:
No complaints from me... but then it was my solution to the issue :)

pmanrubia:
@maxgds sorry for going in circles with this fix. I am merging it.

PR #398, opened 2019-02-26
Fix repetitive name/acronym combinations

maxgds:
# What
Remove repetitive organisation name/acronym combinations such as NS&I (NS&I)  or Companies House (Companies House). This removes extraneous whitespace and then compares the strings to determine whether to display the value from the acronym field.

# Why
Adds noise and no value, so needs removing

# Screenshots
## Before
screen shot 2019-02-26 at 08 50 27
 
## After
screen shot 2019-02-26 at 08 50 10

### DIFF
--- a/app/presenters/filter_presenter.rb
+++ b/app/presenters/filter_presenter.rb
@@ -76,8 +76,11 @@ class FilterPresenter
     @organisation_options ||= begin
       additional_organisation_options +
         @organisations.map do |org|
-          name = org[:name]
-          name.concat " (#{org[:acronym]})" if org[:acronym].present?
+          name = org[:name].strip
+          if org[:acronym].present?
+            acronym = org[:acronym].strip
+            name.concat " (#{acronym})" if acronym != name
+          end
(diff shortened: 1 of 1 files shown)


app/presenters/filter_presenter.rb
```diff
@@ -76,8 +76,11 @@ def organisation_options
     @organisation_options ||= begin
       additional_organisation_options +
         @organisations.map do |org|
-          name = org[:name]
-          name.concat " (#{org[:acronym]})" if org[:acronym].present?
+          name = org[:name].strip
+          if org[:acronym].present?
+            acronym = org[:acronym].strip
+          end
+          name.concat " (#{acronym})" if acronym.present? && acronym != name
```
cbaines: Could the above line be brought inside the `if`? That looks like it would allow removing the `acronym.present?` part of the conditional. The behaviour would mostly be the same, only differing for acronyms that are all whitespace.

maxgds: paired & done

PR #405, opened 2019-03-04
Change colour to de-emphasise response values.

maxgds:
# What
This splits the satisfaction data in two at the presenter level,
adds tests for the correct responses, then presents the section in brackets
in a paler colour than previously (The colour has been checked for contrast/accessibility).
https://trello.com/c/RHEteDvs/1173-1-change-the-colour-of-the-number-of-responses-text-inside-the-table

# Why
De-emphasise response numbers

# Screenshots

## Before
screen shot 2019-03-04 at 09 20 11

## After
screen shot 2019-03-04 at 09 19 06

-

### DIFF
--- a/app/presenters/content_row_presenter.rb
+++ b/app/presenters/content_row_presenter.rb
@@ -2,27 +2,35 @@ class ContentRowPresenter
   include ActionView::Helpers::NumberHelper
   include ActiveSupport::Inflector
 
-  attr_reader :title, :base_path, :document_type, :upviews,
-              :user_satisfaction, :searches, :raw_document_type
+  attr_reader :title, :base_path, :document_type, :upviews, :searches,
+    :raw_document_type, :satisfaction_percentage, :satisfaction_responses
+
   def initialize(data)
     @title = data[:title]
     @base_path = format_base_path(data[:base_path])
     @raw_document_type = data[:document_type]
     @document_type = humanize(data[:document_type])
     @upviews = number_with_delimiter(data[:upviews], delimiter: ',')
-    @user_satisfaction = format_satisfaction(
-      data[:satisfaction], data[:useful_yes], data[:useful_no]
-    )
+    @satisfaction_percentage = format_satisfaction_percentage(data[:satisfaction])
+    @satisfaction_responses = format_satisfaction_responses(data[:useful_yes], data[:useful_no])
+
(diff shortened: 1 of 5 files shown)


app/views/content/index.html.erb
```diff
@@ -134,7 +134,14 @@
...
               <td class="govuk-table__cell govuk-table__cell--numeric" data-gtm-id="upviews-column"><%= item.upviews %></td>
-              <td class="govuk-table__cell govuk-table__cell--shaded" data-gtm-id="satisfaction-column"><%= item.user_satisfaction %></td>
+              <td class="govuk-table__cell govuk-table__cell--shaded" data-gtm-id="satisfaction-column">
+                <%= item.user_satisfaction_percentage %>
+                <% if item.user_satisfaction_responses %>
+                  <span class="satisfaction-column__responses">
+                    (<%= item.user_satisfaction_responses %>)
+                  </span>
+                <% end %>
+              </td>
```
theseanything: Could we re-write the logic like so that we know that `user_satisfaction_percentage` is always a percentage. Currently it can return `No responses` which could be confusing.  
```erb
<% if item.user_satisfaction_percentage %>
    <%= item.user_satisfaction_percentage %>
    <span class="satisfaction-column__responses"> (<%= item.user_satisfaction_responses %>) </span>
<% else %>
   <%= item.user_satisfaction_responses %>
<% end %>
```

app/presenters/content_row_presenter.rb
```diff
@@ -3,26 +3,34 @@ class ContentRowPresenter
...
     )
+
     @searches = number_with_delimiter(data[:searches], delimiter: ',')
   end
 
 private
 
-  def format_satisfaction(score, yes_responses, no_responses)
+  def format_satisfaction_percentage(score)
     return 'No responses' unless score
```
theseanything: Could we have `format_satisfaction_responses` method return this? Currently its confusing what each method should return. Instead this method could be like 
```suggestion
    return "#{(score * 100).round(0)}%" if score
```
and it would work with the suggested changes to the view below.

theseanything:
@maxgds feel free to reach out if things don't make sense.

maxgds:
@theseanything Thanks for the review. I think I've addressed everything so could you please revisit?

PR #407, opened 2019-03-07
Adjust spacing on data page

maxgds:
# What
https://trello.com/c/OUCjbciM/1204-2-spacing-adjustments-for-the-page-data-page
Change spacing on the data page for tidier visuals.
- This wraps the data page in a class so shared styles won't leak (seems to have been lost when it was present previously) and applies adjusted padding and margins to the page. 
- Makes a minor adjustment to the chart size as we no longer need to accommodate the legend so the space at the top is wasted.

# Why
Tidier page.

## Before
screencapture-content-data-admin-publishing-service-gov-uk-metrics-vehicle-tax-2019-03-07-11_04_09

## After
# Screenshots
screencapture-content-data-admin-dev-gov-uk-metrics-vehicle-tax-2019-03-07-10_26_37

### DIFF
--- a/app/assets/stylesheets/components/_chart.scss
+++ b/app/assets/stylesheets/components/_chart.scss
@@ -11,7 +11,7 @@
 
 .app-c-chart__table {
 
-  max-height: 350px;
+  max-height: 300px;
   margin-top: govuk-spacing(3);
 
   @include media(mobile) {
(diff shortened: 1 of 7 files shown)


maxgds:
@alex-ju I'd appreciate you glancing over this as it's all FE code, but it's very basic tweaks so should be a quick one.

alex-ju:
Sure, I'll have a look. I think the PR description is missing the 'after' screenshot.

maxgds:
Extra screenshot added

maxgds:
There's not any actual renaming, it's just gotten a bit confused with being wrapped and compared things wrongly, I think. The actual changes are just the spacing values, so that's cool, thanks.

PR #410, opened 2019-03-08
Add feature tests for tracking clicking app name

theseanything:
# What
This adds a feature test clicking the app name in the header.

# Why
We need to be sure we don't break tracking with Google Analytics.

---
# Review Checklist
* [x] Changes in scope.
* [x] Added/updated unit tests.
* [x] Added to Trello card.

### DIFF
--- a/spec/features/index_page/user_analytics_spec.rb
+++ b/spec/features/index_page/user_analytics_spec.rb
@@ -54,4 +54,8 @@ RSpec.feature 'user analytics' do
   scenario 'tracks time period reveal' do
     expect(page).to have_selector('[data-gtm-id="time-period-options"] summary')
   end
+
+  scenario 'tracks clicks on app name in header' do
+    expect(page).to have_selector('.govuk-phase-banner__content__app-name')
+  end
 end


theseanything:
[(2) Track clicks on 'Content data' app name](https://trello.com/c/N5F9dKYs/1212-2-track-clicks-on-content-data-app-name)

spec/features/index_page/user_analytics_spec.rb
```diff
@@ -54,4 +54,8 @@
   scenario 'tracks time period reveal' do
     expect(page).to have_selector('[data-gtm-id="time-period-options"] summary')
   end
+
+  scenario 'tracks clicks on app name in header' do
```
cbaines: An explicit mention of the tracking being in Google Analytics, or Google Tag Manager (wherever it is) might be good, otherwise someone could see this as a redundant test.

theseanything: All the tests in this spec are for making sure our dependencies in our codebase for analytics and user tracking are present. It'd be weird to be more specific for this particular scenario. Not sure I understand why someone would see it as redundant?

cbaines: Ah, yeah, I just looked at the diff, I didn't actually notice this was a feature test or what file it was in. In that case it's fine :+1:

PR #411, opened 2019-03-12
Move help text into modals

maxgds:
# What

https://trello.com/c/D1sGgrBa/1144-5-view-help-text-in-table-frontend
- Move the help text into modals that appear on click
- Offer a non-js experience (also allows user to right click and open in new window)
- Support testing page elements that rely on js and add a test for modals

# Why
Title text was an interim measure, users have been seen struggling with it in testing as it only appears on hover, and even then is patchy. This is way better for accessibility and general usability.

# Screenshots

## Before
Screen Shot 2019-03-12 at 11 39 37

## After
Screen Shot 2019-03-12 at 11 39 56

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,7 @@ group :development, :test do
   gem "capybara"
   gem "factory_bot_rails"
   gem "govuk-lint", "~> 3"
+  gem "govuk_test"
   gem 'pry'
   gem "rspec-rails", "~> 3"
   gem "spring-commands-rspec"
--- a/spec/features/index_page/index_filter_by_title_spec.rb
+++ b/spec/features/index_page/index_filter_by_title_spec.rb
@@ -13,7 +13,6 @@ RSpec.describe '/content' do
     GDS::SSO.test_user = build(:user, organisation_content_id: 'org-id')
 
     stub_content_page(time_period: 'last-month', organisation_id: 'org-id', items: items)
-
     visit '/content?submitted=true&date_range=last-month&organisation_id=org-id'
   end
 
@@ -36,6 +35,14 @@ RSpec.describe '/content' do
 
     it 'populates the search field with the current filter' do
       expect(page).to have_selector('input[name=search_term][value=title]')
+    end
+
+    it 'launches help text in a modal', js: true do
+      help_string = I18n.t('metrics.upviews.about')
+      expect(page).to_not have_text(help_string)
+      click_link(href: "/help/?hkey=upviews")
+      expect(page).to have_selector('.gem-c-modal-dialogue__box')
+      expect(page).to have_text(help_string)
     end
   end
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -44,6 +44,8 @@ GEM
       tzinfo (~> 1.1)
     addressable (2.6.0)
       public_suffix (>= 2.0.2, < 4.0)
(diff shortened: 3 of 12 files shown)


Gemfile
```diff
@@ -32,8 +32,10 @@ group :development, :test do
   gem "capybara"
   gem "factory_bot_rails"
   gem "govuk-lint", "~> 3"
+  gem "govuk_test"
   gem 'pry'
   gem "rspec-rails", "~> 3"
+  gem "selenium-webdriver"
```
cbaines: I think the `govuk_test` gem requires `selenium-webdriver` so I'm wondering if having this entry in the Gemfile is necessary/useful?

spec/features/index_page/index_filter_by_title_spec.rb
```diff
@@ -37,5 +36,12 @@
     it 'populates the search field with the current filter' do
       expect(page).to have_selector('input[name=search_term][value=title]')
     end
+
+    it 'launches help text in a modal', js: true do
+      expect(page).to_not have_text('This represents the number of visits in which the page was viewed at least once.')
+      click_link(href: "/help/?hkey=upviews")
+      expect(page).to have_selector('.gem-c-modal-dialogue__box')
+      expect(page).to have_text('This represents the number of visits in which the page was viewed at least once.')
+    end
```
cbaines: For future reference, it might be a bit clearer to keep the test changes (this one) separate from the test setup stuff, just so the commits are a bit smaller and more focused.

spec/features/index_page/index_filter_by_title_spec.rb
```diff
@@ -37,5 +36,12 @@
     it 'populates the search field with the current filter' do
       expect(page).to have_selector('input[name=search_term][value=title]')
     end
+
+    it 'launches help text in a modal', js: true do
+      expect(page).to_not have_text('This represents the number of visits in which the page was viewed at least once.')
+      click_link(href: "/help/?hkey=upviews")
+      expect(page).to have_selector('.gem-c-modal-dialogue__box')
+      expect(page).to have_text('This represents the number of visits in which the page was viewed at least once.')
```
cbaines: Maybe extract this string in to a variable, or perhaps reference the locales file somehow, as I think that's how other strings are handled in the tests?

Gemfile
maxgds: Good spot, removed with no ill effect.

spec/features/index_page/index_filter_by_title_spec.rb
maxgds: Fair point - I'm just getting used to multi-commit PRs but there's clearly room for improvement.

spec/features/index_page/index_filter_by_title_spec.rb
maxgds: Fixed

maxgds:
Could you take another look, please @cbaines

spec/features/index_page/index_filter_by_title_spec.rb
cbaines: :100:

spec/features/index_page/index_filter_by_title_spec.rb
cbaines: Great :+1:

PR #418, opened 2019-03-14
Enable ActionMailer library

theseanything:
Enabling ActionMailer as the [app deployment scripts](https://github.com/alphagov/govuk-app-deployment/pull/303) now symlinks ActionMailer configuration in the initialiser. This fixes the issue where ActionMailer configuration cannot find ActionMailer during deployment.


cbaines:
> This will cause deployments to fail.

Are we saying this commit will cause deployments to fail?

theseanything:
> > This will cause deployments to fail.
> 
> Are we saying this commit will cause deployments to fail?

Sorry, that was confusing - updated commit message.

cbaines:
> Sorry, that was confusing - updated commit message.

Thanks, the updated one makes more sense :+1:

PR #424, opened 2019-03-18
Add GTM data attribute to sort links

maxgds:
A final tweak to support the modals as per
https://trello.com/c/D1sGgrBa/1144-5-view-help-text-in-table-frontend

This adds an attribute to make it easier to differentiate between clicks on the sort link and the help link inside the table header.


maxgds:
I'll make a separate ticket to deal with adding the tests and do that tomorrow. Thanks for a prompt review.

PR #429, opened 2019-03-19
Add CSV download via email link

theseanything:
# What
This adds the ability to download CSV of the data on the content page via an link emailed to the user.
Changes include:
- Adding CSV download link on content page
- Add Sidekiq process and worker support
- Add CSV export Sidekiq worker that generate the CSV file, uploads to S3 and emails the user.
- Add ActionMailer to mail the user
- Add download request confirmation page when user clicks on download CSV

# Why
This is to provide CSV format of the data on the content page. Previously the user could download the file by streaming the download as it was generated. However, this had issues with server timeouts and quirks in browsers behaviours around streaming files. 

This approach aims to generate the file before making it available for the user to download. The CSV is generated asynchronously as it is a long blocking process. The file is then uploaded to S3, as the file is too large to attach to the email directly. The user is then sent an email, to alert them that the data is ready to download. This should provide a more stable experience for downloading the data and be more consistent with other apps in GOV.UK that also use this workflow.

# Screenshots

## After
image

---
# Review Checklist
* [x] Changes in scope.
* [x] Added/updated unit tests.
* [x] Added to Trello card.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -76,6 +76,7 @@ GEM
       nokogiri (~> 1.8)
     coderay (1.1.2)
     concurrent-ruby (1.1.5)
+    connection_pool (2.2.2)
     crack (0.4.3)
       safe_yaml (~> 1.0.0)
     crass (1.0.4)
@@ -142,6 +143,13 @@ GEM
       rake
       rouge
       sass-rails (>= 5.0.4)
+    govuk_sidekiq (3.0.3)
+      gds-api-adapters (>= 19.1.0)
+      govuk_app_config (>= 1.1)
+      redis-namespace (~> 1.6)
+      sidekiq (>= 5, < 6)
+      sidekiq-logging-json (~> 0.0)
+      sidekiq-statsd (~> 0.1)
     govuk_test (0.4.1)
       capybara
       chromedriver-helper
@@ -242,6 +250,8 @@ GEM
     rack (2.0.6)
     rack-cache (1.9.0)
       rack (>= 0.4)
+    rack-protection (2.0.5)
+      rack
     rack-test (1.1.0)
       rack (>= 1.0, < 3)
     rails (5.2.2.1)
@@ -274,6 +284,9 @@ GEM
     rb-fsevent (0.10.3)
     rb-inotify (0.10.0)
       ffi (~> 1.0)
+    redis (4.1.0)
+    redis-namespace (1.6.0)
+      redis (>= 3.0.4)
     regexp_parser (1.3.0)
     request_store (1.4.1)
       rack (>= 1.4)
@@ -337,6 +350,17 @@ GEM
       rubyzip (~> 1.2, >= 1.2.2)
     sentry-raven (2.7.4)
       faraday (>= 0.7.6, < 1.0)
+    sidekiq (5.2.5)
+      connection_pool (~> 2.2, >= 2.2.2)
+      rack (>= 1.5.0)
+      rack-protection (>= 1.5.0)
+      redis (>= 3.3.5, < 5)
+    sidekiq-logging-json (0.0.18)
+      sidekiq (>= 3)
+    sidekiq-statsd (0.1.5)
+      activesupport
+      sidekiq (>= 2.6)
+      statsd-ruby (>= 1.1.0)
     simplecov (0.16.1)
       docile (~> 1.1)
       json (>= 1.8, < 3)
@@ -399,6 +423,7 @@ DEPENDENCIES
   govuk-lint (~> 3)
   govuk_app_config (~> 1)
   govuk_publishing_components (~> 16.6)
+  govuk_sidekiq (~> 3)
   govuk_test
   kaminari
   listen (~> 3)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -86,6 +86,7 @@ GEM
     domain_name (0.5.20180417)
       unf (>= 0.0.5, < 1.0.0)
     erubi (1.8.0)
+    excon (0.62.0)
     execjs (2.7.0)
     factory_bot (5.0.0)
       activesupport (>= 4.2.0)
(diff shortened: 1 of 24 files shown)


app/workers/csv_export_worker.rb
```diff
@@ -38,7 +38,7 @@ def upload_to_s3(key, body)
       aws_secret_access_key: ENV['AWS_SECRET_ACCESS_KEY']
     )
 
-    directory = connection.directories.get(ENV['AWS_S3_BUCKET_NAME'])
+    directory = connection.directories.get(ENV['AWS_CSV_EXPORT_BUCKET_NAME'])
```
cbaines: This change, and the changes below would make a bit more sense in a different commit.

app/workers/csv_export_worker.rb
```diff
@@ -2,6 +2,8 @@
 
 class CsvExportWorker
   include Sidekiq::Worker
+  sidekiq_options retry: 0
+  sidekiq_options queue: 'export_csv'
```
cbaines: :+1:

Gemfile.lock
```diff
@@ -113,7 +113,7 @@ GEM
       fog-core
       nokogiri (>= 1.5.11, < 2.0.0)
     formatador (0.2.5)
-    gds-api-adapters (57.4.2)
+    gds-api-adapters (57.5.0)
```
cbaines: As this change has hit master, you can probably just rebase this branch https://github.com/alphagov/content-data-admin/pull/430

cbaines:
This looks pretty good, some changes maybe would be better split across different commits but that's all I've seen so far.

Given that the entire CSV file is held in memory, I'd be interested in knowing how this performs when exporting everything, especially how much memory the Sidekiq process eats.

app/workers/csv_export_worker.rb
```diff
@@ -0,0 +1,57 @@
+require 'fog/aws'
+
+class CsvExportWorker
+  include Sidekiq::Worker
+  sidekiq_options retry: 0
+  sidekiq_options queue: 'export_csv'
```
pmanrubia: What about using `ActiveJob` API to set the queue name and retry options?:

```ruby
retry_on StandardError, attempts: 3
queue_as :default
```

app/workers/csv_export_worker.rb
```diff
@@ -0,0 +1,57 @@
...
+    content_items = FindContent.enum(search_params)
+    ContentItemsCSVPresenter.new(
+      content_items,
+      search_params,
+      document_types,
+      organisations
+    )
+  end
+
+  def upload_to_s3(filename, body)
```
pmanrubia: Uploading a file to S3 seems like a concern that should not live in the job. Why not just extracting a class with this behaviour and reuse it from this class? Testing would also look better.

app/workers/csv_export_worker.rb
```diff
@@ -0,0 +1,57 @@
...
+    basename = presenter.filename
+    csv_string = presenter.csv_rows
+
+    file_url = upload_to_s3("#{basename}.csv", csv_string)
+
+    # Send email with link
+    ContentCsvMailer.content_csv_email(recipient_address, file_url).deliver_now
+  end
+
+  def get_csv_presenter(search_params)
```
pmanrubia: Can we rename this method to `build_csv_presenter`? `get_XXX` feels a bit awkward to me :)

app/workers/csv_export_worker.rb
```diff
@@ -0,0 +1,57 @@
...
+    document_types = FetchDocumentTypes.call[:document_types]
+    organisations = FetchOrganisations.call[:organisations]
+
+    content_items = FindContent.enum(search_params)
```
pmanrubia: Not for this PR, but something that I would address shortly is:

1. Add models `Organisation` and `DocumentType`
2. Add methods to return all organisations and document types with memoization
3. Remove those parameters from the presenter API

This would simplify how we build the presenter.

spec/workers/csv_export_worker_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+
+    allow(FetchDocumentTypes).to receive(:call)
+      .and_return(document_types: document_types)
+    allow(FetchOrganisations).to receive(:call)
+      .and_return(organisations: organisations)
+    allow(FindContent).to receive(:enum)
+      .with(search_params)
+      .and_return(content_items)
+
+    Fog.mock!
```
pmanrubia: As per my comment above we would not need mock all this requests as they can be tested in a unit test.

spec/workers/csv_export_worker_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+
+  let(:organisations) do
+    [{ name: 'org', id: 'org-id', acronym: 'OI', },
+     { name: 'another org', id: 'another-org-id', },
+     { name: 'Users Org', id: 'users-org-id', acronym: 'UOI', }]
+  end
+
+  let(:content_items) do
+    [
+      {
```
pmanrubia: We are setting many values in the test that are not in used. Factories might help here. Have you considered using FactoryBot to simplify these tests? This could also benefit the feature test.

app/workers/csv_export_worker.rb
theseanything: Separated out into separate commit

cbaines: Great, thanks, I'll take a look.

spec/workers/csv_export_worker_spec.rb
theseanything: Agreed - however, I suggest this being a separate PR as we have a few other test cases where the FactoryBot could be used. It would take a bit effort to refactor this.

spec/workers/csv_export_worker_spec.rb
theseanything: Still need to mock this for the worker as will ultimately call S3.

app/workers/csv_export_worker.rb
theseanything: I've moved to being a concern

theseanything:
@pmanrubia @cbaines I've addressed you comments!

spec/workers/csv_export_worker_spec.rb
pmanrubia: @theseanything: Sorry if I wasn't clear on my previous message: in the unit test you could mock the method of the class that you have extracted (you own the API), considering that:

1. The unit test for the class that you have extracted would do the testing
2. The feature test would not mock anything

PR #435, opened 2019-03-26
Abbreviate glance metric figures

maxgds:
# What
https://trello.com/c/aWmUQbVq/947-2-fix-overflowing-figures-on-glance-metrics
This updates the Glance Metric component to let it abbreviate the figures it is passed and apply the visible label and screenreader support label. It updates the existing tests and adds a couple more.

# Why
Glance metrics were written to take input of an abbreviated number and the appropriate labels, but we have never passed in the data in that format. This means that the text is often stretching the limits of its container and occasionally overflows.

# Screenshots

Note that the before/after are not using the same dataset - the code does properly round the numbers.

## Before
Screen Shot 2019-03-26 at 08 33 41

## After
Screen Shot 2019-03-26 at 08 33 50

### DIFF
--- a/app/views/components/_glance-metric.html.erb
+++ b/app/views/components/_glance-metric.html.erb
@@ -1,16 +1,30 @@
 <%
+  measurement_display_labels = {Thousand: 'k', Million: 'm', Billion: 'b'}
   name ||= false
   figure ||= false
   period ||= false
-  measurement_display_label ||= ""
-  measurement_explicit_label ||= ""
   context ||= ""
   trend_percentage ||= ""
+
+  if figure
+    number = number_to_human(figure.gsub(/,/, '').to_i, :format => '%n', :precision => 3)
+    measurement_explicit_label = number_to_human(figure.gsub(/,/, '').to_i, :format => '%u', :precision => 3)
+    if figure.end_with? "%"
+      measurement_display_label = "%"
+    else
+      measurement_display_label = measurement_display_labels[measurement_explicit_label.to_sym]
+    end
+  end
 %>
 <% if name && figure && period %>
   <div class="app-c-glance-metric govuk-body">
     <h3 class="app-c-glance-metric__heading"><%= name %></h3>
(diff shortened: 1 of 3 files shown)


cbaines:
Putting the motivation for the change in the Pull Request is helpful, but putting it in the commit message would also be helpful.

maxgds:
> putting it in the commit message would also be helpful.

Added.

app/views/components/_glance-metric.html.erb
```diff
@@ -1,16 +1,30 @@
...
+
+  if figure
+    number = number_to_human(figure.gsub(/,/, '').to_i, :format => '%n', :precision => 3)
+    measurement_explicit_label = number_to_human(figure.gsub(/,/, '').to_i, :format => '%u', :precision => 3)
+    if figure.end_with? "%"
+      measurement_display_label = "%"
+    else
+      measurement_display_label = measurement_display_labels[measurement_explicit_label.to_sym]
+    end
+  end
```
cbaines: If possible, it would be good to move this code in to a presenter class, as that's more consistent with how the formatting for numbers is done currently. Maybe this one https://github.com/alphagov/content-data-admin/blob/master/app/presenters/single_content_item_presenter.rb , although I haven't fully looked at where the values are coming from, or what's happening to them.

maxgds:
@alex-ju what are your thoughts on the level of logic this component is handling? Chris was suggesting that we should do the calculations elsewhere, but this allows us to re-use the string that is produced and fed to other parts of the page. My feeling is that it's relatively lightweight and it's not unusual for a component to apply formatting at this level but if you agree with Chris I can rework it.

app/views/components/_glance-metric.html.erb
alex-ju: I think @cbaines has a good point that this should be considered 'complex' display logic, hence in a Rails world it should sit in a presenter so that it can potentially be shared between views. However, coming from a non-Rails world, I struggle to provide guidance on how it would ideally look like. I did it myself for the table component and tried to extract some bits in govuk_publishing_components and proved to be good practice.

maxgds: Cheers, both. I'm going to retire this PR and will raise another one with amendments later.

PR #454, opened 2019-04-01
Split filter and content presenter

pmanrubia:
[Trello card](https://github.com/alphagov/content-data-admin/pull/454)

### What

1. It moves out `FilterPresenter` from `ContentItemsPresenter`
2. It inlines code to retrieve `Organisations` and `DocumentTypes`
3. It simplifes code and tests to make them more readable

**Note**

In next PR we need to move into `ActiveModel` the code related to `Organisations` and `DocumentTypes` as we should not handle JSON directly within the application.

### Code sample

```ruby
  def index
    document_types = FetchDocumentTypes.call[:document_types]
    organisations = FetchOrganisations.call[:organisations]

    search_results = FindContent.call(search_params)

    @presenter = ContentItemsPresenter.new(
      search_results, search_params, document_types, organisations,
    )
  end
```

to:

```ruby
  def index
    search_results = FindContent.call(search_params)

    @filter = FilterPresenter.new(search_params)
    @presenter = ContentItemsPresenter.new(search_results, search_params)
  end
```

### DIFF
--- a/spec/presenters/filter_presenter_spec.rb
+++ b/spec/presenters/filter_presenter_spec.rb
@@ -17,12 +17,13 @@ RSpec.describe FilterPresenter do
     }
   end
 
+  before do
+    allow(FetchDocumentTypes).to receive(:call).and_return(document_types: document_types)
+    allow(FetchOrganisations).to receive(:call).and_return(organisations: organisations)
+  end
+
   subject do
-    FilterPresenter.new(
-      search_parameters,
-      document_types,
-      organisations,
-    )
+    FilterPresenter.new search_parameters
   end
 
   describe '#document_type_options' do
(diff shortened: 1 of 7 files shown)


spec/presenters/filter_presenter_spec.rb
```diff
@@ -17,12 +17,13 @@
...
+    allow(FetchOrganisations).to receive(:call).and_return(organisations: organisations)
+  end
+
   subject do
-    FilterPresenter.new(
-      search_parameters,
-      document_types,
-      organisations,
-    )
+    FilterPresenter.new search_parameters
```
theseanything:
```suggestion
    FilterPresenter.new(search_parameters)
```

PR #474, opened 2019-04-15
Support `local_env.yml`

peteglondon:
# What
Support local environment variables for `development` and `test`
environments, defined in `config/local_env.yml`.

This is the same as `content-performance-manager` which supports
it in `development`.

The file is git-ignored so we can all have different ones.

# Why

Local (mac not VM) development is difficult.

This enables you to have local environment variables.
This is very useful if, for instance, you want your database to be configured in 
a non-standard way (i.e. in Docker)

You don't have to use it but it makes life easier if you do.

---
# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@
 .byebug_history
 
 /coverage
+/config/local_env.yml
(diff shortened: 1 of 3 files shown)


cbaines:
> This enables you to have local environment variables.

Have you tried using something like https://direnv.net/ perhaps?

Also, the database configuration looks to use `DATABASE_URL` from the environment, I'd be interested to know what more is required?

cbaines:
Maybe I should have rejected this, rather than just commenting, but I don't like the extra configuration added here, especially without any specific reason for why doing this is helpful. so I'd be interested in continuing the conversation...

PR #490, opened 2019-04-24
Update content data link when exporting CSVs

MahmudH:
- An update has been made to reflect the new link
- content-data replaces content-data-admin

Trello: https://trello.com/c/mr3p168v

### DIFF
--- a/spec/presenters/content_items_csv_presenter_spec.rb
+++ b/spec/presenters/content_items_csv_presenter_spec.rb
@@ -119,7 +119,7 @@ RSpec.describe ContentItemsCSVPresenter do
       end
 
       it 'has a Content Data Link' do
-        expect(subject[3]).to start_with('http')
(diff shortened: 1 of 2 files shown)


spec/presenters/content_items_csv_presenter_spec.rb
```diff
@@ -122,6 +122,10 @@
         expect(subject[3]).to start_with('http')
       end
 
+      it 'has correct Content Data Link' do
+        expect(subject[3]).to eq('http://content-data.dev.gov.uk/metrics')
+      end
```
cbaines: How about extending the scope of the preceding test, rather than adding a new one, as they seem quite similar?

MahmudH: Sounds good, I'll update that now 👍

cbaines: Great :+1:

PR #501, opened 2019-04-30
Show comma separator in comparison data

MahmudH:
- Add delimiter to comparison data

# Screenshots
Screen Shot 2019-04-30 at 17 18 42

The comparison data does not show a delimiter, so this PR fixes this.

### DIFF
--- a/spec/support/response_helpers.rb
+++ b/spec/support/response_helpers.rb
@@ -196,11 +196,11 @@ module GdsApi
             },
             {
               name: "searches",
-              total: 48,
+              total: 6,
               time_series: [
-                { "date" => day1, "value" => 16 },
-                { "date" => day2, "value" => 16 },
-                { "date" => day3, "value" => 16 }
+                { "date" => day1, "value" => 2 },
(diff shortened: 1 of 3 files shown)


spec/support/response_helpers.rb
```diff
@@ -196,11 +196,11 @@ def older_single_page_response(base_path, from, to)
             },
             {
               name: "searches",
-              total: 48,
+              total: 6,
               time_series: [
-                { "date" => day1, "value" => 16 },
-                { "date" => day2, "value" => 16 },
-                { "date" => day3, "value" => 16 }
```
cbaines: I don't quite follow why these numbers are changing?

MahmudH: The trend percentage is calculated with these metric values. I wanted to make the percentage of increase higher than +406.25% to be in the thousands. Decreasing these numbers increases the percentage of increase so replaced these to 2. The 2 was picked randomly and did give a good result, 3,950.

cbaines: Ok, thanks :+1:

PR #527, opened 2019-05-23
Improve layout of page data page which has no data to show

MahmudH:
When a content is published there is no data present until the ETL process
executes. Currently, the page does not hint that there is no data present,
this may confuse users to think that the page maybe broken.

Now the page is styled like other pages for consistency and displays the sentence,
no data, to make it more clear.

Trello card: https://trello.com/c/JOfeWXgS

# Screenshots

## Before
Screen Shot 2019-05-23 at 15 57 14

## After
Screen Shot 2019-05-23 at 15 55 31

### DIFF
--- a/app/views/components/_glance-metric.html.erb
+++ b/app/views/components/_glance-metric.html.erb
@@ -1,20 +1,25 @@
 <%
   name ||= false
-  figure ||= false
+  figure ||= nil
   period ||= false
   context ||= ""
   trend_percentage ||= ""
   measurement_display_label ||= ""
   measurement_explicit_label ||= ""
 %>
-<% if name && figure && period %>
+<% if name && period %>
   <div class="app-c-glance-metric govuk-body">
     <h3 class="app-c-glance-metric__heading"><%= name %></h3>
-    <span class="app-c-glance-metric__figure"><%= figure %><% if measurement_display_label %><span class="app-c-glance-metric__measurement"
-        <% if measurement_explicit_label %>
-          aria-label="<%= measurement_explicit_label %>"
-        <% end %>><%= measurement_display_label %></span>
-    <% end %>
+
+      <% if figure.nil? %>
+      <span class="app-c-glance-metric__figure">No data</span>
+      <% else %>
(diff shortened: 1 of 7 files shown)


theseanything:
Just adding a note to say that the satisfaction score still shows as a 0% because of a bug in the backend api where it returns 0 instead of nil.

theseanything:
@maxgds maybe good if you could review this too?

spec/components/glance_metric_spec.rb
```diff
@@ -23,8 +23,9 @@
   end
 
   it "does not render if figure is not supplied" do
```
maxgds: Please update the test to say it renders "no data" as it used to render nothing and if this is failing it will be misleading trying to figure out what to fix.

spec/components/info_metric_spec.rb
```diff
@@ -23,8 +23,9 @@
   end
 
   it "does not render if figure is not supplied" do
```
maxgds: As above - make sure your test description matches exactly what you expect to see.

app/views/components/_glance-metric.html.erb
```diff
@@ -1,20 +1,25 @@
...
   <div class="app-c-glance-metric govuk-body">
     <h3 class="app-c-glance-metric__heading"><%= name %></h3>
-    <span class="app-c-glance-metric__figure"><%= figure %><% if measurement_display_label %><span class="app-c-glance-metric__measurement"
-        <% if measurement_explicit_label %>
-          aria-label="<%= measurement_explicit_label %>"
-        <% end %>><%= measurement_display_label %></span>
-    <% end %>
+
+      <% if figure.nil? %>
+      <span class="app-c-glance-metric__figure">No data</span>
```
maxgds: I might have added the class name to app-c-glance-metric__figure--no-data and added a rule in the CSS to reduce the size of the font to make it a little tidier, but I'm not super-bothered about this, seeing this version of the page at all is going to be an edge case and I'm not sure it's worth adding to the CSS for it.

spec/components/glance_metric_spec.rb
MahmudH: Good call, I've updated both tests 👍

PR #557, opened 2019-06-07
Add link to edit in content publisher.

peteglondon:
# What
Adds a link `Edit in Content publisher` when the `publishing_app`
is `content-publisher`.

What are the proposed changes?

# Why
As a convenience to users

---
# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.

### DIFF
--- a/spec/helpers/external_links_helper_spec.rb
+++ b/spec/helpers/external_links_helper_spec.rb
@@ -20,6 +20,20 @@ RSpec.describe ExternalLinksHelper do
       end
     end
 
+    context 'with content-publisher' do
+      it 'generates a link to the content-publisher app' do
+        expect(
+          edit_url_for(content_id: 'content_id',
+            publishing_app: 'content-publisher',
+            base_path: '/anything',
+            document_type: 'news_story',
+            locale: 'fr')
+        ).to eq(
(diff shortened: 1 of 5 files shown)


peteglondon:
[(1) Add link to edit content in Content Publisher](https://trello.com/c/emVzo7OQ/1446-1-add-link-to-edit-content-in-content-publisher)

spec/helpers/external_links_helper_spec.rb
```diff
@@ -141,6 +155,16 @@
         )
       end
     end
+
+    context 'with content-publisher' do
+      it 'uses the specific label' do
+        expect(
+          edit_label_for(publishing_app: 'content-publisher')
+        ).to eq(
+          'Edit in Content publisher'
```
theseanything: Can we use the locales in the test? - `I18n.t('...')`
```

peteglondon: I don't see the advantage looking at all the other tests in the file. Also we are testing the 
translations kook up correctly.

theseanything: So the reasoning was that if we do make content changes, content designers can just make them and not have to worry about the tests.

peteglondon: Done now

PR #573, opened 2019-06-19
Compare

maxgds:
# What & why
Build the frontend side of the compare page and links. This is currently split into 4 commits

- the compare page itself
- the link from the actions box at the top of the page
- the description of parts and the updated header in the Document Content section of the page
- a link to the compare page from the document section of the metrics page

There were some small linting issues in the original commit that are fixed in the subsequent one, but it's otherwise good to be used independently. There are no tests for the compare page (commit 1), and its route needs to change. 

Commits 2 and 3 have tests against them although some of the logic provided by the helper in commit 3 may not be needed, depending on how much the api provides. Commit 4 relies on the helper in commit 3.


theseanything:
Closing this PR as doesn't need to be merged.

PR #577, opened 2019-06-20
Add compare page

maxgds:
Things needed to make this production ready:
- Update routes (and filenames?)
- Proper page titles (the html title)
- Data to populate the page header (visible title and kicker)
- Chunks of comparison data to populate the table
- Make the sorts in the table headers respond properly
- Pass in relevant wording for the table caption (not all multi-part content is sections)  - could use helper from here: https://github.com/alphagov/content-data-admin/pull/573/commits/8a74aaede5671553461e39ce50b414561b54f8eb
- Translation strings where relevant


spec/features/document_children_page/documents_children_page_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    xit 'links to the page data page after filtering' do
+      click_on 'The title'
+      expect(page).to have_content("Page data: #{I18n.t('metrics.show.time_periods.last-month.leading')}")
+    end
+
+    xit 'selected organisation is shown in dropdown menu' do
+      expect(page).to have_select('organisation_id', selected: 'another org')
+    end
+
```
peteglondon: Should all these tests be ignored?

peteglondon: ^^ Ignore that - I see they get sorted in a later commit

PR #583, opened 2019-06-24
Fix sentence case for document types in table summary

maxgds:
# What
https://trello.com/c/DmEHFLCa/1514-capitalisation-of-acronyms-in-summary-above-index-page-table

We previously made a conscious decision to downcase document type names, but this has led to items with abbreviated terms or acronyms coming out incorrectly, eg html instead of HTML. This PR adds logic to the presenter to exclude the relevant terms from being downcased.

# Screenshots

## Before
Screen Shot 2019-06-24 at 11 52 28

## After
Screen Shot 2019-06-24 at 11 52 38

### DIFF
--- a/app/views/content/_table_data_description.html.erb
+++ b/app/views/content/_table_data_description.html.erb
@@ -9,7 +9,7 @@
         for '<span class="table-caption__param"><%= filter.search_terms %></span>'
     <% end %>
     <% if filter.document_type? && filter.document_type_id != "all" %>
-        for <span class="table-caption__param"><%= filter.document_type_name.downcase %></span>
+        for <span class="table-caption__param"><%= filter.document_type_name.humanize(capitalize: false) %></span>
     <% else %>
         for <span class="table-caption__param">all document types</span>
     <% end %>


cbaines:
Have you got an example of something that needs downcasing?

maxgds:
> Have you got an example of something that needs downcasing?

Anything that doesn't contain an abbreviation, eg:

Screen Shot 2019-06-24 at 13 22 05

cbaines:
> > Have you got an example of something that needs downcasing?
> 
> Anything that doesn't contain an abbreviation, eg:

Right, ok, so the Content Data API returns the name with the first letter capitalised.

It uses the `humanize` method to do this [1], and that method does have a `capitalize` option, which you can set to false [2].

1: https://github.com/alphagov/content-data-api/blob/1566525634faf7397e435e6679c73f2ebe39690f/app/domain/finders/all_document_types.rb#L11
2: https://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-humanize

If you call the `humanize` method in the Content Data Admin with the `capitalize: false` option, I think it'll give you the string you want.

```ruby
irb(main):001:0> 'About our services'.humanize(capitalize: false)
=> "about our services"
irb(main):002:0> 'AAIB report'.humanize(capitalize: false)
=> "AAIB report"
```

maxgds:
Thanks, @cbaines - fixed as per your suggestion and I've learned a new method! Could you please re-review?

PR #591, opened 2019-06-25
Fix table row presenter with nil data

theseanything:
This fixes a bug where the table row presenter doesn't know how to process nils for useful_yes, useful_no, satisfaction or searches. Now it will return the string 'No data'.

---
# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.

### DIFF
--- a/app/presenters/content_row_presenter.rb
+++ b/app/presenters/content_row_presenter.rb
@@ -12,11 +12,11 @@ class ContentRowPresenter
     @raw_document_type = data[:document_type]
     @document_type = humanize(data[:document_type])
     @sibling_order = data[:sibling_order] || '-'
-    @upviews = number_with_delimiter(data[:upviews], delimiter: ',')
+    @upviews = number_with_delimiter(data[:upviews], delimiter: ',') || 'No data'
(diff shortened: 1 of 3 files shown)


app/presenters/content_row_presenter.rb
```diff
@@ -26,6 +26,8 @@ def format_satisfaction_percentage(score)
   end
 
   def format_satisfaction_responses(yes_responses, no_responses)
+    return 'No data' if yes_responses.nil? || no_responses.nil?
+
```
maxgds: Possibly stupid question... but if 1 person has responded, or all responses are the same should we really be returning no data?

theseanything: Yeah so it be very weird to for that backend to return say `1` yes_responses and `nil` no_responses. But in that case because we still have a `nil` and we can't calculate the percentage.

In this context `nil` is not the same as 0, `nil` means we don't know the value i.e. it could be 0 or it could 1000 etc. And without knowing exactly the number of both yes and no response we can't calculate the percentage.

PR #708, opened 2019-12-20
Disable showing comparison data

cbaines:
As it could be misleading now that the way data is collected through
Google Analytics has changed. Also tweak the CSS to adjust for the
removal of the comparison data.

# Screenshots

## Before
homepage-before

## After
image

---
# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.

### DIFF
--- a/app/views/components/_glance-metric.html.erb
+++ b/app/views/components/_glance-metric.html.erb
@@ -22,31 +22,5 @@
       <% end %>
     </span>
     <p class="app-c-glance-metric__context govuk-body-xs"><%= context %></p>
-    <span class="app-c-glance-metric__trend">
-      <% if trend_percentage.blank? %>
-        no comparison data
-      <% else %>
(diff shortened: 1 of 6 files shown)


app/views/components/_glance-metric.html.erb
```diff
@@ -22,31 +22,5 @@
       <% end %>
     </span>
     <p class="app-c-glance-metric__context govuk-body-xs"><%= context %></p>
-    <span class="app-c-glance-metric__trend">
-      <% if trend_percentage.blank? %>
```
bilbof: Could `trend_percentage` be removed too, or is this needed elsewhere?

cbaines: Good point, I'm hoping that this is a temporary change so I've actually tried to keep as much of the code in as possible, I ended up just removing this block though.

Hopefully it can be decided soon whether to properly remove this as a feature, or make it reliably work again!

PR #846, opened 2020-09-18
Upgrade ruby 2.7

benbaumann95:
This PR contains the changes necessary for the app to run on Ruby 2.7.1.

The version of Ruby is left the same, since there is a script that can be run after all the apps have been updated to update the version all at once

- Tested locally with ruby 2.7.1 and 2.6.6
- Upgrade capybara
- Fix kwargs deprecation warnings


spec/features/date_selection_spec.rb
```diff
@@ -18,9 +18,13 @@
   context "no date period provided" do
     it "renders the correct date ranges in the time select" do
       visit page_uri
+
+      Capybara.ignore_hidden_elements = false
       time_labels = find(".app-c-time-select").all(".govuk-radios__item").map do |el|
         { el.find("label").text => el.find("span").text }
       end
+      Capybara.ignore_hidden_elements = true
+
```
cdccollins, line 27: What's going on here?

benbaumann95: Have explained in: https://github.com/alphagov/content-data-admin/pull/846/commits/9699f322cd305b9f2f4a2ec371b8b80d707cfa56

Usually would just set this globally but there are other specs that rely on not showing hidden elements. Happy for a better suggestion though.

cdccollins: Ah lol sorry, hate when people don't look at my commit messages 😂

cdccollins (approved):

PR #864, opened 2020-10-23
Pin govuk_publishing_components to 17.21.0

alex-ju:
Content Data uses the accessible autocomplete component which was removed in version 18.
Arose from https://github.com/alphagov/govuk-developer-docs/pull/2830

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ gem "fog-aws"
 gem "gds-api-adapters"
 gem "gds-sso"
 gem "govuk_app_config"
-gem "govuk_publishing_components"
+gem "govuk_publishing_components", "~> 17" # Content Data uses the accessible autocomplete component which was removed in version 18
 gem "govuk_sidekiq"
 gem "kaminari"
 gem "mail-notify"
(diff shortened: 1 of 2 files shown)


theseanything:
Ah eek! We probably had a good reason to remove autocomplete - I'm a bit worried that Content Data, is now stuck with an old version and will stop receiving any other publishing component updates/fixes. Do we plan to remove autocomplete from Content Data? Or provide autocomplete in another way?

alex-ju:
> Ah eek! We probably had a good reason to remove autocomplete - I'm a bit worried that Content Data is now stuck with an old version and will stop receiving any other publishing component updates/fixes. Do we plan to remove autocomplete from Content Data? Or provide autocomplete in another way?

We removed the component as it was JavaScript-heavy and, at that time consuming individual components from `govuk_publishing_components` was not an option. Given the current configuration allows this, we can bring the component back in the gem but in theory, we shouldn't bring a component to the gem unless there's a need to share it ( 🤷‍♂️ ). So an alternative would be to bring the component in `content-data-admin`. I'm personally more leaning towards having it in the gem, as it keeps it uncoupled from any application. I'll create a card for this bit of frontend tech debt to keep track of it. Do let me know if you prefer one way or another – I tagged you and Kevin because you were tech leads on this at certain points.

injms (approved):

PR #942, opened 2021-04-26
Remove legacy /healthcheck route

barrucadu:
govuk-puppet and govuk-aws have now been updated to only use the new
routes, so the old one is no longer needed.

See RFC 141 for more information.

---

[Trello card](https://trello.com/c/tl6RV1el/723-improve-govuk-healthchecks)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,13 +1,6 @@
 Rails.application.routes.draw do
   root to: redirect("/content")
 
-  get "/healthcheck", to: proc { [200, {}, %w[OK]] }
-  get "/healthcheck-f",
-      to: proc {
-            GovukError.notify("Sentry works")
-            [200, {}, ["Sentry notified"]]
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -1,13 +1,6 @@
 Rails.application.routes.draw do
   root to: redirect("/content")
 
-  get "/healthcheck", to: proc { [200, {}, %w[OK]] }
-  get "/healthcheck-f",
-      to: proc {
-            GovukError.notify("Sentry works")
-            [200, {}, ["Sentry notified"]]
-          }
```
AlanGabbianelli, line 9: Do we have/need an alternative to this?

barrucadu: As far as I can tell it's not checked by anything

AlanGabbianelli (approved):

PR #993, opened 2021-12-07
Use Postgres 13 in CI 

1pretz1:
Points the app to run against Postgres version 13, instead of the
old version 9.6. Postgres 13 runs on the port 54313.

Trello:

* https://trello.com/c/QQf4LjT6/14-content-data-admin

* https://trello.com/c/yaBLaC1O/957-upgrade-content-data-admin-postgres-96-13

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,6 +2,10 @@
 
 library("govuk")
 
-node("postgresql-9.6") {
+node {
+
+  // Run against the Postgres 13 Docker instance on GOV.UK CI
+  govuk.setEnvar("TEST_DATABASE_URL", "postgresql://postgres@127.0.0.1:54313/content-data-admin-test")
+
   govuk.buildProject()
 }


config/environments/development.rb
```diff
@@ -64,10 +64,6 @@
   # Annotate rendered view with file names.
   # config.action_view.annotate_rendered_view_with_filenames = true
 
-  # Use an evented file watcher to asynchronously detect changes in source code,
-  # routes, locales, etc. This feature depends on the listen gem.
-  config.file_watcher = ActiveSupport::EventedFileUpdateChecker
```
hannako, line 69: So is this just tidying up to remove an error, that must have been around since 7 May! Rather than anything to do with the postgres upgrade? If so I'd be inclined to do that in a fresh branch and PR. Incase everything goes wonky and we end up reverting.

1pretz1: Correct, it shows how much Content Data is worked on 😆  I was being a bit lazy including it in this PR so I'll split them into two.

1pretz1: Done!

hannako (approved):
LGTM!

PR #1076, opened 2022-08-08
Don't use ES6 syntax in survey-banner script

andyhd:
Deploy task fails because of ES6 syntax.


richardTowers (approved):
Looks good to me.

Arrow functions => ordinary functions is fine because there's no use of `this`.

Template string to string concatenation also looks exactly equivalent.

PR #1104, opened 2022-10-05
Fix Javascript error when survey banner turned off

andyhd:
The script that handles showing and hiding the survey banner breaks on pages other than the index page, because it assumes the element is present.

This fix simply checks that the element is present before attempting to show or hide it.


maxf (approved):
Looking good, thanks.

sengi (approved):

PR #1175, opened 2023-02-03
Upgrade govuk_publishing_components and Plek to current versions.

sengi:
Add the accessible-autocomplete component directly, since it was removed from govuk_publishing_components in version 18. This is the upgrade path recommended in the [govuk_publishing_components changelog](https://github.com/alphagov/govuk_publishing_components/blob/main/CHANGELOG.md#1800).

This unblocks upgrading from the now very outdated version 17 of govuk_publishing_components, which unblocks upgrading to the current version of [Plek].

Removed the no-longer-necessary version constraints and ran `bundle update` to update to govuk_publishing_components 34.7.1 and Plek 5.0.0 (plus minor updates to some other gems).

[Plek]: https://github.com/alphagov/plek

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,8 @@ FROM $builder_image AS builder
 WORKDIR $APP_HOME
 COPY Gemfile* .ruby-version ./
 RUN bundle install
+COPY package.json yarn.lock ./
+RUN yarn install --production --frozen-lockfile --non-interactive --link-duplicates
 COPY . .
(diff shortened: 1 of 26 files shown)


MuriloDalRi:
@sengi I had not seen this PR until today but I think we did this work [here](https://github.com/alphagov/content-data-admin/pull/1183)

Can we close this one?

sengi:
Yeah I never quite got around to getting this working in time unfortunately! Oh well :)

PR #1183, opened 2023-03-29
Move code for accessible autocomplete component into app

MuriloDalRi:
This component was removed in version 18 of govuk_publishing_components https://github.com/alphagov/govuk_publishing_components/pull/1038

We need to move the code into the app so that we can update govuk_publishing_components and other dependencies.

https://trello.com/c/egaF1kjo/3110-replace-accessible-autocomplete-for-content-data-admin


app/assets/javascripts/components/accessible-autocomplete.js
```diff
@@ -0,0 +1,67 @@
...
+      }
+    }
+  }
+
+  function track (category, action, label, options) {
+    if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
+      options = options || {}
+      options.label = label
+
+      window.GOVUK.analytics.trackEvent(category, action, options)
```
andysellick, line 62: I think this is a hangover from the original code - it's calling the Universal Analytics code for Google Analytics tracking (in the components gem). I believe content-data-admin uses GTM for tracking now, so this is probably not needed? I don't know what the tracking requirements for this component are, so it might be out of the scope of this PR to implement it here.

andysellick: If UA tracking is not required I think this entire function and the `if` statement on line 36 can be removed as well.

app/assets/stylesheets/components/_accessible-autocomplete.scss
```diff
@@ -0,0 +1,35 @@
...
+.app-c-accessible-autocomplete {
+  .autocomplete__input {
+    z-index: 1;
+  }
+
+  .autocomplete__dropdown-arrow-down {
+    z-index: 0;
+  }
+
+  .autocomplete__option {
```
andysellick, line 12: I think some additional styles might be needed to fix a problem with the first, blank option in the autocomplete, which is shown below as the dark blue highlighted strip.

Screenshot 2023-04-17 at 10 30 46

As you can see it doesn't look like a selectable option, but it  allows users to clear the selection in the autocomplete by clicking, rather than having to delete the text manually if required. I remember this from the first implementation, it should be fixable with something like `min-height: 1em;`, which would result in the following:

Screenshot 2023-04-17 at 10 34 07

spec/javascripts/components/accessible-autocomplete-spec.js
```diff
@@ -0,0 +1,126 @@
+/* eslint-env jasmine, jquery */
+/* global GOVUK */
+
+describe('An accessible autocomplete component', function () {
+  'use strict'
+
+  function loadAutocompleteComponent () {
+    window.setFixtures(html)
+    var autocomplete = new GOVUK.Modules.AccessibleAutocomplete()
+    autocomplete.start($('.app-c-accessible-autocomplete'))
```
andysellick, line 10: This needs updating in line with the changes to the start/init/jQuery/etc. stuff. I think it should be something like:

```JavaScript
var el = document.querySelector('.app-c-accessible-autocomplete')
var autocomplete = new GOVUK.Modules.AccessibleAutocomplete(el)
autocomplete.init()
```

spec/javascripts/components/accessible-autocomplete-spec.js
```diff
@@ -0,0 +1,126 @@
...
+        done()
+      })
+    })
+
+    it('the input is cleared', function () {
+      expect($('select').val()).toEqual('')
+    })
+  })
+
+  describe('triggers a Google Analytics event', function () {
```
andysellick, line 75: Tests for Google Analytics might not be needed if `content-data-admin` is using GTM for all tracking, see my earlier comment.

andysellick (commented):
Looks good so far, needs a couple of minor changes.

app/assets/javascripts/components/accessible-autocomplete.js
robinjam: Hey Andy, thanks for the review 🙏 

If we want to learn more about the tracking requirements for this app, who should we speak with? Or should we just remove the tracking from this component entirely and consider it out of scope for this PR?

robinjam: I've done some more digging and it looks like we aren't tracking interactions with this component anywhere in this app:

The tracking is only set up if `data-track-category` and `data-track-action` are both set on the `<select>` element. These attributes would be set in `app/views/components/_accessible_autocomplete.html.erb` if a `data_attributes` local is passed in. I've had a look through all the views where we include this partial, and none of them pass any locals when rendering the partial.

Given this, I'm going to remove the tracking code, which (I think) also allows us to remove the entire `onConfirm` function (as the remainder of the function just seems to be a workaround for a bug caused by the fact that we override `onConfirm`.

I'll try to summarise this in the commit message.

spec/javascripts/components/accessible-autocomplete-spec.js
robinjam: Good catch, also need to update the CI so that these tests actually get run 😅

app/assets/javascripts/components/accessible-autocomplete.js
andysellick: Sorry for the slow response, yes, I think removing the tracking code entirely is the right approach. Removing the `onConfirm` also cleans up the code nicely, although can you check that the hidden `select` element is still updated when the user selects something in the autocomplete?

spec/javascripts/components/accessible-autocomplete-spec.js
andysellick: I was wondering why there weren't any errors! 😁

andysellick (approved):
Changes look good 👍 

Might be worth seeing if any of the commits could be squashed together, but not a blocker.

PR #1235, opened 2023-07-04
Site improve report

KludgeKML:
Adds SiteImprove calls into Content Data Admin metrics pages

https://trello.com/c/kam9jbQF/27-check-content-for-accessibility-issues-against-siteimprove-api

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #1246, opened 2023-07-13
Add content quality information from Siteimprove to page data

leenagupte:

Attempts to tidy up: https://github.com/alphagov/content-data-admin/pull/1235
Trello: https://trello.com/c/kam9jbQF

# What's changed?

Use the [Siteimprove API] to get the content and accessibility issues about a page on GOV.UK and display alongside the other page data. 

This information will only currently display for mainstream content for two reasons:

1. The current licence for Siteimprove only allows for 10,000 documents to be analysed
2. Mainstream content is the only content that has been added to the "site" (the list of pages to be evaluated) on Siteimprove.

# Why?

Current ways to find content and accessibility issues have resulted in a lot of toil. 

Content designers are already using Siteimprove, which provides many things about of the box to help reduce this toil. 
Siteimprove provides a dashboard to allow pages to be manually to be evaluated, and provides an [html report] with all of the issues. A Siteimprove account is not required to view the page report for an individual page. Custom policies can also be added to Siteimprove that allow the setting of bespoke content quality rules.

However, there is still toil in getting the link to the report and sending it to the publisher. 

Siteimprove has an API that provides descriptions of the issues and direct links to specific parts of the html report. Adding the data from the API to Content Data Admin will allow publishers to see data about their content in one place, could be expanded to be used for all GOV.UK content.

# Expected changes

screenshot-content-data integration publishing service gov uk-2023 07 06-18_33_57

# Next steps
- [ ] Each page is currently added to Siteimprove manually. The pages should be able to be added programmatically from the [sitemap] in Publisher.
- [ ] Add the Siteimprove data about the page to the data warehouse. This will allow content to be sorted by the number of content issues on the index page.
- [x] Publish the [siteimprove_api_client] gem.

[Siteimprove API]: https://api.siteimprove.com/v2/documentation#/
[html report]: https://my2.siteimprove.com/Inspector/1054012/A11Y/Page?pageId=47277004219&impmd=0D32B24C60EE55FEAF4B62ADB33FEC69&conf=a+aa+aria+si#/sia-r78/failed/RXNOMnEybVM=;ZDgzOWI2ZTM=
[sitemap]: https://publisher.publishing.service.gov.uk/govuk-sitemap.xml
[siteimprove_api_client]: https://github.com/alphagov/siteimprove_api_client

Co-authored-by: Keith Lawrence <keith.lawrence@digital.cabinet-office.gov.uk>
Co-authored-by: Max Harden <max.harden@digital.cabinet-office.gov.uk>


app/presenters/siteimprove/policy_issues_presenter.rb
```diff
@@ -0,0 +1,53 @@
...
+          policy_direct_link: "#{link}#issue/#{issue.id}",
+        }
+      end
+    end
+
+    def link
+      summary_info._siteimprove.policy.page_report.href
+    end
+
+    def policy_type?(policy_name)
```
robinjam, line 33: I think the `?` in this method's name should be dropped, as it doesn't return a boolean

(I assume it used to return a boolean in your spike and it just hasn't been renamed since then?)

robinjam (approved):
This is wonderful, amazing work everyone!

Thanks for cleaning up the commit history, and making the commit descriptions so detailed, it made reviewing it a breeze.

I note that you’re quite heavily using `Rails.cache`, which is a great idea, but I’m also conscious that this app is configured to use Rails’ default in-memory cache store in prod. This has a limit of 32MB by default, is not shared between web pods (each pod gets its own separate cache), and will reset every time a new version of the app gets deployed. That might not be an issue, but let me know if it is and I we can look into moving to a real cache store.

One small code change suggestion, but it's not a blocker, so LGTM 👍

PR #1329, opened 2023-12-19
Make page size an env var

MuriloDalRi:
This is now defined in govuk helm charts: https://github.com/alphagov/govuk-helm-charts/pull/1636

https://trello.com/c/MU9KwkRP/3368-improve-performance-of-content-data-csv-exports-3


AgaDufrat (approved):
Nice!

PR #1402, opened 2024-03-21
Update the GA4 banner with details about inconsistencies

AgaDufrat:
Some issues with the data were reported. We don't think we should email users via Notify, because we can't tell how many of the potential 4,599 users are actually actively using the app (access to the app is given by default). This message will only be seen by users who are actually using the tool.


nicholsj (approved):
Thanks, LGTM!

PR #1430, opened 2024-04-26
Add distinct label to CSV export metrics

deborahchua:
Prometheus scrapes data from Pushgateway every 60 seconds which means it continues to record the same metric until another one is sent. Adding a distinct label will allow us to filter out duplicate values in Grafana.

As you can only register a histogram once, we have to create a new one which allows us to add a label as part of it's setup.

Related PR: https://github.com/alphagov/content-data-admin/pull/1426
Trello card: https://trello.com/c/SQro2G8f/3476-monitor-how-long-content-datas-csv-exports-take-5


AgaDufrat (approved):
LGTM. I wonder if we can deleted the histogram and set it up again to avoid this name change.

PR #1450, opened 2024-05-17
Removing link to ua

PeterHattyar:
Removed a link to UA because we are migrating to GA4. This link is now redundant.

Before: 
BeforeLinkRemoval
After:
AfterLinkRemoval

https://trello.com/c/QacJuXBT/3490-remove-a-link-to-universal-analytics-in-content-data-1

### DIFF
--- a/config/locales/defaults/en.yml
+++ b/config/locales/defaults/en.yml
@@ -65,7 +65,7 @@ en:
       context: '%{percent_users_searched}% of users searched from the page'
       unit: 'search terms'
       data_source: google_analytics
-      external_link: 'See search terms in Google Analytics (opens in new tab)'
+      external_link: ''
       about_title: 'About searches from page'
       about: >
         This is the number of internal site searches that were started from the page. When people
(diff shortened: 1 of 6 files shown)


config/locales/defaults/en.yml
```diff
@@ -65,7 +65,7 @@ en:
       context: '%{percent_users_searched}% of users searched from the page'
       unit: 'search terms'
       data_source: google_analytics
-      external_link: 'See search terms in Google Analytics (opens in new tab)'
```
AgaDufrat, line 68: We could technically remove this definition entirely now, as it's not used anywhere. It was used in
https://github.com/alphagov/content-data-admin/blob/main/app/views/metrics/_metric_section.html.erb#L17 but since external_link in `nil` it won't get rendered. However there's a pattern of leaving things blank in this file so let keep it as it is.

AgaDufrat (approved):
Looks good to me. Thank you for including screenshots. I left a comment but it's ok to merge it like that

PR #1565, opened 2024-10-08
Configure hosts for application

brucebolt:
Note: the healthcheck endpoints are requested by IP, not domain, so we need to specifically exclude them from the protection.<br><br>[Trello card](https://trello.com/c/frqFN7D8)

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -85,10 +85,10 @@ Rails.application.configure do
   }
 
   # Enable DNS rebinding protection and other `Host` header attacks.
-  # config.hosts = [
-  #   "example.com",     # Allow requests from example.com
-  #   /.*\.example\.com/ # Allow requests from subdomains like `www.example.com`
-  # ]
+  config.hosts = [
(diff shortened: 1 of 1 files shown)


AgaDufrat (approved):
LGTM

brucebolt:
I've realised this allows domains like `app-name.dodgygov.uk`.  The regex will need to be updated to `app-name\..*\.gov.uk$`.

nicholsj:
In case it's relevant, note [the additional allowed hostname for local-links-manager](https://github.com/alphagov/local-links-manager/pull/1432), because the original restriction caused issues

AgaDufrat:
The author discontinued work on this PR. Reverting to Draft (so it's not pick up by Seal) and adding it to our team's backlog.

It requires further work:
- add internal hostname to allowed domains as the API is used internally in the cluster by argo workflows for deployments 
- the regex is too permissive and will need updating to `/appname\..*\.gov.uk$/`

dj-maisy:
@brucebolt - Closing this as well. Pleas re-open if this is progressed.

PR #1609, opened 2024-11-21
Upgrade app to Rails 7.2

MuriloDalRi:
Upgrades Rails from 7.1 to 7.2

https://github.com/alphagov/govuk-infrastructure/issues/1444


config/environments/test.rb
```diff
@@ -34,13 +32,19 @@
...
   config.action_mailer.perform_caching = false
 
   # Tell Action Mailer not to deliver emails to the real world.
   # The :test delivery method accumulates sent emails in the
   # ActionMailer::Base.deliveries array.
   config.action_mailer.delivery_method = :test
 
+  # Unlike controllers, the mailer instance doesn't have any context about the
+  # incoming request so you'll need to provide the :host parameter yourself.
+  config.action_mailer.default_url_options = { host: "www.example.com" }
```
AgaDufrat, line 46:
```suggestion
config.action_mailer.default_url_options = { host: Plek.find("content-data", external: true) }
```

AgaDufrat (approved):
Left a comment. 

I think we want to use our own error pages templates (not the rails defaults). I'd personally remove the icons but may be best to check what approach was taken for the the frontend/rendering apps and follow what they did.

PR #2047, opened 2026-03-31
(CM-835) Update accessibility guidance links

graycodes:
The old publishing guidance will be removed and replaced with these new pages.

We should wait until the new guidance is live before merging this.

https://gov-uk.atlassian.net/browse/CM-835

https://govuk.zendesk.com/agent/tickets/6442185


pezholio (approved):
These links seem to be live now, so approving

PR #2174, opened 2026-08-12
Upgrade js-yaml to 5.2.3 to fix vulnerability

edavey:
This addresses a vulnerability in js-yaml (see [CVE-2026-59870][cve]).

There are 3 consumers of js-yaml, requiring the following. All are now pulled up to v 5.2.3:

```
"@eslint/eslintrc@npm:^0.3.0":
  js-yaml: ^3.13.1

"cosmiconfig@npm:^9.0.2":
    js-yaml: ^4.1.0

"eslint@npm:~7.18.0":
    js-yaml: ^3.13.1
```

See https://github.com/alphagov/content-data-admin/security/dependabot/208

[cve]:
https://github.com/advisories/GHSA-724g-mxrg-4qvm

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,8 +35,7 @@
     "accessible-autocomplete": "^3.0.1"
   },
   "resolutions": {
-    "js-yaml@^3.13.1": "^3.14.2",
-    "js-yaml@^4.1.0": "^4.3.0"
+    "js-yaml": "^5.2.2"
   },
   "packageManager": "yarn@3.5.0"
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -312,15 +312,6 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -35,8 +35,7 @@
     "accessible-autocomplete": "^3.0.1"
   },
   "resolutions": {
-    "js-yaml@^3.13.1": "^3.14.2",
```
patrickjfl, line 38: hmmm, that's interesting. I thought that we were tied to a v3 resolution because of our dependence on standardx (which is no longer maintained), which uses eslint v7, which uses js-yaml v3. But I've checked out the branch and `npm run lint:js` seems to be working fine. I guess there have been breaking changes to js-yaml since v3 but thankfully they haven't broken anything for eslint v7?

patrickjfl (approved):

